PR #14, opened 2017-03-29
updated urls for HTTP API Level 0

rhoads-zach:
per https://github.com/obsidian-toaster/documentation/wiki/HTTP-API-Level-0-Experience

### DIFF
--- a/docs/topics/experience-deploy-quickstart.adoc
+++ b/docs/topics/experience-deploy-quickstart.adoc
@@ -30,5 +30,5 @@ Once your quickstart is deployed and started, you can check its route informatio
 ----
 $ oc get routes
 NAME                 HOST/PORT                                         PATH      SERVICES             PORT      TERMINATION
-app1                 app1.192.168.99.100.nio.io                        app1                           8080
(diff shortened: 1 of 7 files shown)


rhoads-zach:
@inoxx03 please take a look and let me know what you think

rhoads-zach:
@cescoffier mind taking a quick look?

ALRubinger:
@rhoads-zach "api" namespace changes look good to me.

rhoads-zach:
thank you @cescoffier i'm still seeing "id" in the quick starts for boot and wf-swarm. is that incorrect?

ALRubinger:
99bddf0

heiko-braun:
I think all urls should refer to `nip.io`, rather than `nio.io`

ALRubinger:
Good catch @heiko-braun

rhoads-zach:
nice catch @heiko-braun 

I updated the PR to use the correct urls

Mind taking a quick look and providing a thumbs up if it looks good?

heiko-braun:
Looks good to me

ALRubinger:
f736639a9ef0d55b74c5406ab21897b25c5f0988

PR #52, opened 2017-04-07
first pass at getting started

rhoads-zach:
Related to #51


rhoads-zach:
@inoxx03 or @ALRubinger please take a quick look if you have time

PR #55, opened 2017-04-09
Fix #54 - Add health check

### LINKED ISSUE #54: Add Vert.x Health Check for CI
Continuation of work done in #53 

We could use a Health Check to ensure rolling deployment doesn't roll until the verticle is ready to serve.  Is this not put in place by FMP?

cescoffier:
Also fixes tests.


ALRubinger:
246489e

PR #65, opened 2017-04-11
updated getting started guide per issue #56

rhoads-zach:


rhoads-zach:
@inoxx03 or @ALRubinger mind taking a look?

Issue #66, opened 2017-04-11
scaffold landing page


rhoads-zach:

rhoads-zach:
@ALRubinger let me know if you need anything else for this.

@inoxx03 mind giving this a +1 if this looks good?

inoxx03:
+1 on the PR, LGTM:)

PR #67, opened 2017-04-11
Issue #66: added scaffold for landing page

rhoads-zach:


ALRubinger:
@rhoads-zach Approved

PR #74, opened 2017-04-12
Add mission-secured-wf-swarm and common secured-mission docs

starksm64:
Here is the first pass at incorporating the common secured mission documents and the specific mission-secured-wf-swarm


rhoads-zach:
nice work @starksm64! I'll merge this and we can take an editing pass after its merged.

Issue #49, opened 2017-04-07
Create link to docs home on all docs


rhoads-zach:

ALRubinger:
Likely need link to home at top too?  Footer merged cfb528c

PR #89, opened 2017-04-17
Issue #49 and intro updates

rhoads-zach:
Added CC 4.0 License, intro updates, and Home link at the top


rhoads-zach:
@ALRubinger mind taking a quick look?

Issue #41, opened 2017-04-06
Review and finalize integration details with WF Swarm missions


rhoads-zach:
https://github.com/openshiftio/appdev-documentation/blob/master/docs/topics/rest-level-0-experience-basic-interaction-wf-swarm.adoc

https://github.com/openshiftio/appdev-documentation/blob/master/docs/topics/health-check-experience-basic-interaction-wf-swarm.adoc

https://github.com/openshiftio/appdev-documentation/blob/master/docs/topics/configmap-experience-basic-interaction-wf-swarm.adoc

rhoads-zach:
@heiko-braun mind taking this one?

heiko-braun:
sure, i'll review it later on

heiko-braun:
A few things:

- rest-level 0: the curl commands are wrong ad the payload changed. take a look here for the most recent examples: https://github.com/wildfly-swarm-openshiftio-boosters/wfswarm-rest-http

- health-check: works completely different and is now aligned with the mission contents. take a look here: https://github.com/wildfly-swarm-openshiftio-boosters/wfswarm-health-check

- configmap: minor changes to the pod names, etc. but overall reflects the contents more or less: See  https://github.com/wildfly-swarm-openshiftio-boosters/wfswarm-configmap

heiko-braun:
@rhoads-zach this put's the ball back in your court

rhoads-zach:
thank you @heiko-braun 

@kenfinnigan would you like to update this in the docs, or would you like us to?

kenfinnigan:
If you're able to @rhoads-zach that would be awesome!

rhoads-zach:
#91 and #90 help address some of the issues.

I'm still not clear on how to interact with the health check Mission. @kenfinnigan or @heiko-braun can you issue a PR updating the WF Health Check Mission on how you're supposed to interact with it?

heiko-braun:
@rhoads-zach Sure, I can do that. But I am not clear what you expect. The basic interactions are described in https://docs.google.com/document/d/10ZzCVrxzifQeASAYlgFWERqgbYGj4ucJwCzpb37jTyw/edit#heading=h.we4w2qryfz43 and shared between all runtimes. What's the proper place to put common descriptions?

heiko-braun:
@rhoads-zach I've update both the instructions for deployment as well interaction with booster: https://github.com/openshiftio/appdev-documentation/pull/92

Let me know if this is what you had in mind.

[...]

PR #92, opened 2017-04-18
Issue #41: Update healthcheck instructions for interaction and deployment

heiko-braun:


heiko-braun:
Here's the fix for the basic http booster: https://github.com/openshiftio/appdev-documentation/pull/93

Issue #94, opened 2017-04-18
Vert.x HTTP mission issues


cescoffier:
1. The oc pods output is not as presented:

```
$ oc get pods -w
NAME                             READY     STATUS      RESTARTS   AGE
vertx-rest-1-aaaaa       1/1       Running     0          58s
vertx-rest-s2i-1-build   0/1       Completed   0          2m
```
But is actually:

```
NAME                     READY     STATUS      RESTARTS   AGE
http-vertx-1-qzttx       1/1       Running     0          4m
http-vertx-s2i-1-build   0/1       Completed   0          4m
```

The name has changed.

2. The oc get route output has changed:

New output:

```
NAME         HOST/PORT                                   PATH      SERVICES     PORT      TERMINATION   WILDCARD
http-vertx   http-vertx-myproject.192.168.64.12.nip.io             http-vertx   8080                    None
```
3. Service interactions have changed:

```
$ curl http://vertx-rest-myproject.192.168.99.100.nip.io/greeting
{
  "content" : "Hello, World!"
}
```

should be:

```
$ curl http://vertx-rest-myproject.192.168.99.100.nip.io/api/greeting
{
  "content" : "Hello, World!"
}
```

Same issue for the call with `name=sarah`

PR #103, opened 2017-04-18
Isssue #94: updated app name, url, and spacing issue

rhoads-zach:


rhoads-zach:
@cescoffier mind reviewing and giving a +1?

Issue #98, opened 2017-04-18
Vert.x Config Map mission issues


cescoffier:
1. Missing instruction: `oc policy add-role-to-user view -n $(oc project -q) -z default` when creating the project

2. As in other booster, the pod must not only be running but also ready

3. These instructions should be removed:

```
Rollout the new version of your application so the ConfigMap configuraiton changes are picked up.

$ oc rollout latest dc/vertx-http-configmap

Check the status of your booster and ensure your new pod is running.

$ oc get pods -w
NAME                             READY     STATUS      RESTARTS   AGE
vertx-http-configmap-1-aaaaa       1/1       Running     0          58s
vertx-http-configmap-s2i-1-build   0/1       Completed   0          2m
```

Again, vert.x picks the change automatically.

PR #104, opened 2017-04-18
Issue #98: updated as per feedback in issue

rhoads-zach:


rhoads-zach:
@cescoffier mind reviewing and giving me a +1?

PR #110, opened 2017-04-19
Add script to install launchpad mission control on minishift

cmoulliard:
Goal of the script :

- Deploy Launchpad mission control template using the parameters passed to authenticate the user using user/password or token on minishift,
- Setup the Github identity (account & token)
- Patch jenkins to use admin as role


cmoulliard:
@rhoads-zach Why did you close my PR ?

Issue #122, opened 2017-04-19
Secured Rest booster instructions should be runtime specific


heiko-braun:
Currently the instructions combine docs on various runtimes. It would be better for each to reflect just the specific runtime it deals with.

starksm64:
There is a common component as a section related to the specific runtime. Are you talking about the example output?

heiko-braun:
@starksm64 Maybe my assessment is wrong. Did you intentionally reference other runtimes on each <runtime>-booster instruction to showcase the cross runtime interaction?

heiko-braun:
I am specifically talking about the `Interact with the <runtime> Booster` section. It contains parts that don't belong to the current `<runtime>`. I.e. in the swarm docs I see 
![screen shot 2017-04-19 at 18 55 32](https://cloud.githubusercontent.com/assets/369380/25191994/cb1b4120-2531-11e7-872b-01eb26c8eb48.png)

starksm64:
Ok, your right. That was originally written thinking that the java client could be used to access more than one secured booster deployment, but I can see that this could be confusing.

heiko-braun:
I can see arguments for both approaches. At the end it's still a cross functional aspect that we demonstrate. But to support that we would need to make it more explicit how to retrieve and build the other runtime services. 

Thinking of it this may actually complicate things quite a bit for the user, so my suggestion would to constrain the docs to a single runtime in use at a time.

starksm64:
Agreed, I'll change the interaction examples to be distinct for each runtime.

rhoads-zach:
@starksm64 @heiko-braun PR has been merged. I'll take an editing pass on the secured booster guides. If this meets your needs, do either of you mind closing this issue?

heiko-braun:
i've checked and it looks good to me

PR #127, opened 2017-04-19
Secured Rest booster instructions should be runtime specific #122

starksm64:
PR for #122


rhoads-zach:
Looks good overall. Will take an editing pass after the merge.

PR #147, opened 2017-04-21
Fixes for paths in Vert.x config map booster

SlavoKrupa:
@cescoffier  Can you please review ?


cescoffier:
looks good, thanks !

Issue #187, opened 2017-04-26
weird behavior of code snippets when mouse over


Ladicek:
I'm looking at http://appdev-stage.openshift.io/docs/getting-started.html, chapter _2.3.1. Update Using Continuous Delivery_. I see that item _1. Clone your project from GitHub_ has a shell snippet that I can copy and paste, so I try to select the snippet by mouse... but when I move the cursor over the snippet, the word `ruby` appears at the beginning of the line. This happens for a lot of other snippets as well. This happens both in Firefox 53.0 and Chrome 58.0.

rhoads-zach:
I believe this is related to language specified in the source block and the behavior of the syntax highlighter. It seems like it defaults to ruby if no language is specified. I'll try and get to this later today.

Ladicek:
Oh, one clarification: I don't really care what word appears on the beginning of the line. `ruby` certainly looks weird for a shell script, but I'd very much prefer if it didn't happen at all. Sorry if that was obvious, I figured I'd rather clarify :-)

PR #210, opened 2017-04-27
Issue #187: add 'bash' as type for source blocks

rhoads-zach:
@inoxx03 @Ladicek mind taking  a look?

### DIFF
--- a/docs/topics/secured-mission-common-interaction.adoc
+++ b/docs/topics/secured-mission-common-interaction.adoc
@@ -9,10 +9,10 @@ token returned by the authentication step.
 The Java client is packaged as an executable jar and is located at target/sso-client.jar.
 To run the client, use `java -jar target/sso-client.jar -app ...`.
 The full set of command arguments that you can specify can be seen by passing -help:
-+
-[source,options="nowrap",subs="attributes+"]
+
+[source,bash,options="nowrap",subs="attributes+"]
 ----
-[sstark sso]$ java -jar target/sso-client.jar --help
+$ java -jar target/sso-client.jar --help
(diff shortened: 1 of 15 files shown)


Ladicek:
+1 I think it's correct and it's the right thing to do, but it won't fix the original problem, will it?

docs/topics/secured-mission-common-interaction.adoc
```diff
@@ -10,7 +10,7 @@ The Java client is packaged as an executable jar and is located at target/sso-cl
 To run the client, use `java -jar target/sso-client.jar -app ...`.
 The full set of command arguments that you can specify can be seen by passing -help:
 +
-[source,options="nowrap",subs="attributes+"]
+[source,bash,options="nowrap",subs="attributes+"]
 ----
 [sstark sso]$ java -jar target/sso-client.jar --help
```
inoxx03: @rhoads-zach Maybe we should remove sstark's username from the snippet here. WDYT?

rhoads-zach:
@Ladicek I built it locally and the language no longer appears when you hover your mouse over it box, so i believe it does address the initial issue.

Ladicek:
Ah, OK, good!

docs/topics/secured-mission-common-interaction.adoc
rhoads-zach: @inoxx03 fixed

rhoads-zach:
merged: https://github.com/openshiftio/appdev-documentation/commit/4dff338726be15aaf82544f9abfe4dfc571f35ef

PR #225, opened 2017-05-09
Getting started guide: made exmaple URLs non-clickable

### LINKED ISSUE #212: Example partial URL's should not be rendered as clickable links
Partial example URLs in the Getting Started guide  are rendered as clickable links.
These URLs are not supposed to work, so clicking on them only gives you a `Page Not Found` browser screen.
Not a big problem, but it only takes a simple markup fix to solve.
And it would be nice not to have developers click them by mistake and navigate to an error page, which could be distracting.

Instead of this: 
`http://github.com/USERNAME/PROJECTNAME/settings/hooks/`

We could use this:
`+++http://github.com/USERNAME/PROJECTNAME/settings/hooks/+++`

inoxx03:
Fixes #212

### DIFF
--- a/docs/topics/getting-started-launchpad-update-booster-cd.adoc
+++ b/docs/topics/getting-started-launchpad-update-booster-cd.adoc
@@ -1,7 +1,7 @@
 [[update-cd]]
 = Update Using Continuous Delivery
 
-This approach allows you to deploy updates directly from the GitHub repository created by the {launcher}. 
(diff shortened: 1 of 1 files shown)


rhoads-zach:
LGTM.

rhoads-zach:
Merged: https://github.com/openshiftio/appdev-documentation/commit/9d309b4b937bb979c193f359fc94fc747e70ec37

PR #227, opened 2017-05-10
Issue #222 document health check log messages

### LINKED ISSUE #222: Document messages from Health Check
hc messages

That attached image shows the messages generated in OpenShift when the "Kill Me" button is pushed in the Health Check mission.  It may be useful to document these as expected.

inoxx03:
Fixes #222


inoxx03:
@rdruss Please take a look at the changes and let me known if there are any additions or corrections to make. Is this much detail OK, or should we also add a table explaining each message type?

rdruss:
@inoxx03, the content you have added looks fine.  Two things I would note:

1) You can also get to the messages/events by going through the Monitoring tab on the OpenShift console.  I'm not sure you need to document every way to get there, just something I thought I would note.

2) I think it would be useful to mention in the log-output doc [1] that the Unhealthy messages (warnings) listed are expected.  Tbh, the first Unhealthy message is the one I expected.  That last one I believe is more of a timing issue, but I'm not 100% positive.

[1] docs/topics/health-check-mission-web-console-log-output.adoc

rhoads-zach:
Merged: https://github.com/openshiftio/appdev-documentation/commit/d023a6d244609d9edbfc539e426ab9fb3a5db43c

PR #234, opened 2017-05-12
clarify the wording about JPA and ORM

Ladicek:
JPA is a specification that certain ORMs adhere to (e.g. Hibernate
or EclipseLink). These ORMs typically also provide their own public
APIs that can be accessed directly. So the relationship between
the JPA and ORM terms is nuanced: they are not quite the same,
but they are not completely unrelated either. This commit tries
to improve the wording in this area.

Also there are some other small fixes in the CRUD mission docs.

### DIFF
--- a/docs/topics/crud-mission-design-tradeoffs.adoc
+++ b/docs/topics/crud-mission-design-tradeoffs.adoc
@@ -2,7 +2,7 @@
 [width="100%",options="header"]
 |====================================================================
 |Pros           |Cons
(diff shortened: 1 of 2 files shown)


inoxx03:
LGTM. Merging.

inoxx03:
Merged:https://github.com/openshiftio/appdev-documentation/commit/b60b459df8f7a1f7bbf1b6df98458cfa05569e41

PR #242, opened 2017-05-17
Cleanup: add oso content

rhoads-zach:
@ALRubinger mind taking a look?


rhoads-zach:
@inoxx03 mind taking a look?

PR #253, opened 2017-06-01
Issue #223: added oc-env step for minishift

rhoads-zach:
@cmoulliard mind taking a look?

### DIFF
--- a/docs/topics/minishift-install-prereq.adoc
+++ b/docs/topics/minishift-install-prereq.adoc
@@ -13,6 +13,20 @@ $ minishift version
 Minishift version: 1.0.0
 ----
 
+. Determine the command to add the correct version of the `oc` binary to your path and run the command. 
++
+.Example Result of `oc-env`
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ minishift oc-env
+export PATH="/Users/john/.minishift/cache/oc/v1.5.0:$PATH"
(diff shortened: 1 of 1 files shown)


docs/topics/minishift-install-prereq.adoc
```diff
@@ -13,6 +13,19 @@ $ minishift version
 Minishift version: 1.0.0
 ----
 
+. Determine the command to install the correct version of the `oc` binary and run the command.
```
Ladicek: This doesn't really _install_ `oc`, it just adds correct path to $PATH. I'd suggest rewording to something like "Use `minishift oc-env` to add the correct `oc` binary to your path."

(The following item already uses the expression "your path", so I figured it should be enough. If a link to explanation is needed, this Wikipedia page is good: https://en.wikipedia.org/wiki/PATH_(variable) :-) )

Ladicek: Actually with this paragraph added, the WARNING below (line 40) seems no longer necessary, or it could be shortened and moved to this section?

rhoads-zach:
@Ladicek updated per feedback. Let me know what you think, otherwise, i'll merge later today

rhoads-zach:
Thanks @Ladicek, merging.

rhoads-zach:
resolves #223

PR #267, opened 2017-06-08
Update: added first draft of sb runtime details

rhoads-zach:
@cmoulliard @inoxx03 mind taking a look?

### DIFF
--- a/docs/spring-boot-runtime/master.adoc
+++ b/docs/spring-boot-runtime/master.adoc
@@ -7,9 +7,7 @@ include::topics/templates/document-attributes.adoc[]
 
 This guide covers concepts as well as practical details needed by developers to use the {SpringBoot} runtime.
 
-
-== Runtime Details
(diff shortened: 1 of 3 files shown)


Ladicek:
CC @RadekKoubsky

cmoulliard:
> taking a look?

Looks good to me @rhoads-zach . We should perhaps  for all the runtimes include a paragraph where we will explain the added value to run the different runtimes on OpenShift, what should be changed within an existing project to transform it into a compatible project deployable on Openshift (= f-m-p plugin, ....), what will happen during the build (= target folder zipped and pushed to openshift, s2i build started using the Java S2I image to copy/paster and customize the docker image created, create an imagestream which is next used with the DeploymentConfig to install the application

inoxx03:
merging

inoxx03:
merged: 5338ba695e4b2047764c0fb86e9c998bdc84ff79

PR #310, opened 2017-06-28
Removed hard-coded path and made Maven use the local user's cache

quintesse:


quintesse:
@rhoads-zach this is the fix I made to determine the needed path instead of hard-coding it.

I also made the docker image use the local user's Maven cache so things start a lot faster.

Hope you think this useful!

rhoads-zach:
Thank you @quintesse.

LGTM, merging.

rhoads-zach:
merged: https://github.com/openshiftio/appdev-documentation/commit/5b876d9f6b3ce5889c356679a6eecb97d29971ff

PR #312, opened 2017-06-28
Fixes #311 Update Segment tracker id in the doc footer depending on env var

### LINKED ISSUE #311: Update Segment tracker id in the doc footer depending on env var
Righty now the tracker id for analytics is hard-coded, it should use the env var `LAUNCHPAD_TRACKER_SEGMENT_TOKEN` instead

quintesse:


rhoads-zach:
merged: https://github.com/openshiftio/appdev-documentation/commit/54e0524177182efbb17aeb65bd98c62e5c1f3209

quintesse:
Bah it doesn't work (yet):

    analytics.load("");

I added the env var to the staging jenkins openshift app, but it seems it didn't get to where I hoped it would :/

quintesse:
We might need something like described here in the JenkinsFile: https://jenkins.io/doc/pipeline/tour/environment/ but I haven't been able to figure it out yet. Perhaps @gastaldi or @ALRubinger have any ideas?

PR #323, opened 2017-07-01
JBDS (DevStudio) and CDK terminology fixes

rkratky:
Changes:

* 'JBDS' to 'Developer Studio'
* 'the CDK' to 'CDK'
* minor related fixes

See https://mojo.redhat.com/docs/DOC-1088906 for reference.


Ladicek:
Some places use just "Developer Studio", which is a bit too generic. Should maybe use a macro to expand to the full form?

rkratky:
@Ladicek, I believe it's enough to use the full form in the beginning of a section (so the context is established) and then continue with "Developer Studio". Using the full form everywhere would be a mouthful...

Ladicek:
@rkratky I'd agree, but I don't think this is what this PR achieves. It looks like "JBDS" was just replaced by "Developer Studio", so there will be places where "Developer Studio" will occur without context (though I'd have to check the rendered form to be sure about this). Also I think we're just using full forms of everything everywhere in these docs.

rkratky:
You're right. I'll update the PR.

inoxx03:
LGTM:) Please let me know when you update and I'll merge.

rhoads-zach:
@rkratky @inoxx03 are you all still working on this?

rkratky:
@rhoads-zach, I was out for a week. Still in my TODO.

rhoads-zach:
@rkratky do you think you'll get a chance to work on this again in the near term?

inoxx03:
@rkratky If you're busy right now, I can take care of this PR, if that's OK.

rkratky:
@inoxx03, that would be most kind of you. Apologies for stalling on this.

inoxx03:
@Ladicek @rkratky I've changed the name to full form in the title. Admittedly, it does seem a bit longish now, but IMHO this way the following instances of the short form 'Developer Studio' don't appear out of context.

inoxx03:
I'll merge the changes from a parallel branch on my own fork...

inoxx03:
merged: 32def8cc9fdd4607a28790ddc520d518a91c566b

rkratky:
@inoxx03, thanks!

PR #327, opened 2017-07-03
Externalised launcher template URL

### LINKED ISSUE #142: Externalize latest template version; add to release train instructions to update this
Introduced by d8506d013cffafc073861be8864a1b2aafb8f7f9; externalize the URL used by Vert.x as the indirection to `latest`

tradej:
Resolves #142.

In this arrangement, the launcher URL is stored in the `ci/openshiftio-appdev-docs/src/main/resources/application.properties` file under the `launcher.template_url.latest` property. I externalised the `INDEX_PAGE` property to `launcher.index_page` the same way.

If you think the property should be defined in the `pom.xml` file instead, no problem, I'll change it. I am not sure about all the use cases of this product, so I just chose the approach that seemed most bulletproof.

### DIFF
--- a/ci/openshiftio-appdev-docs/src/main/java/io/openshift/appdev/documentation/builder/HttpApplication.java
+++ b/ci/openshiftio-appdev-docs/src/main/java/io/openshift/appdev/documentation/builder/HttpApplication.java
@@ -22,10 +22,35 @@ import io.vertx.ext.web.Router;
 import io.vertx.ext.web.handler.RedirectAuthHandler;
 import io.vertx.ext.web.handler.StaticHandler;
 
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Properties;
+
 public class HttpApplication extends AbstractVerticle {
 
-   private static final String INDEX_PAGE = "index.html";
-   private static final String LAUNCHER_TEMPLATE_LATEST_URL = "https://raw.githubusercontent.com/openshiftio/launchpad-templates/v5/openshift/launchpad-template.yaml";
+    private static final String propFileLocation = "src/main/resources/application.properties";
+    private static final Properties props = loadProperties();
+
+    private static Properties loadProperties() {
+        Properties props = new Properties();
+        FileInputStream  ioStream = null;
+        try {
+            ioStream = new FileInputStream(propFileLocation);
+            props.load(ioStream);
+        } catch (IOException e) {
+            e.printStackTrace();
+        } finally {
+            if (ioStream != null) {
+                try {
+                    ioStream.close();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+        return props;
+    }
+
   @Override
   public void start(Future<Void> future) {
     // Create a router object.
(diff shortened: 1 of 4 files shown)


ci/openshiftio-appdev-docs/src/main/java/io/openshift/appdev/documentation/builder/HttpApplication.java
```diff
@@ -22,26 +22,51 @@
...
-   private static final String INDEX_PAGE = "index.html";
-   private static final String LAUNCHER_TEMPLATE_LATEST_URL = "https://raw.githubusercontent.com/openshiftio/launchpad-templates/v5/openshift/launchpad-template.yaml";
+    private static final String propFileLocation = "src/main/resources/application.properties";
+    private static final Properties props = loadProperties();
+
+    private static Properties loadProperties() {
+        Properties props = new Properties();
+        FileInputStream  ioStream = null;
+        try {
+            ioStream = new FileInputStream(propFileLocation);
```
Ladicek: If it's in `src/main/resources`, it will end up in the application JAR and could be loaded from the classpath I think?

tradej: So, would the loading command then be this?
```java
props.load(HttpApplication.class.getClassLoader().getResourceAsStream("application.properties"));
```

Ladicek: Right, that should be it. Or `HttpApplication.class.getResourceAsStream("/application.properties")`, which is the same.

The benefit is that like this, you don't expect the application sources to exist, the application is self-contained.

tradej: Done, thanks.

ci/openshiftio-appdev-docs/src/main/java/io/openshift/appdev/documentation/builder/HttpApplication.java
```diff
@@ -22,26 +22,40 @@
...
 
-   private static final String INDEX_PAGE = "index.html";
-   private static final String LAUNCHER_TEMPLATE_LATEST_URL = "https://raw.githubusercontent.com/openshiftio/launchpad-templates/v5/openshift/launchpad-template.yaml";
+    private static final String propFileLocation = "/application.properties";
+    private static final Properties props = loadProperties();
+
+    private static Properties loadProperties() {
+        Properties props = new Properties();
+        try {
+            props.load(HttpApplication.class.getResourceAsStream(propFileLocation));
```
Ladicek: You could also lookup both configurables from props and store them into static variables here, but I don't think that's a big deal. Just an idea.

tradej:
I have updated the release train document with instructions to update the URL as well. Please review that document too.

inoxx03:
LGTM:) Would you please squash your commits when you're done making edits? Thanks:)

ci/openshiftio-appdev-docs/src/main/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+launcher.template_url.latest=https\://raw.githubusercontent.com/openshiftio/launchpad-templates/v5/openshift/launchpad-template.yaml
```
rhoads-zach: FYI @tradej you had a regression on the link and i fixed it. It was going to v4 instead of v5. Other than that, this LGTM

rhoads-zach:
Merging.

rhoads-zach:
merged: https://github.com/openshiftio/appdev-documentation/commit/2c726a62c7195847ec7b8efb3c4c305a47aefc4c

ci/openshiftio-appdev-docs/src/main/resources/application.properties
tradej: Thanks for catching that. I don't know how that ended up in there. I forgot to rebase, probably.

PR #350, opened 2017-07-14
Script to tag release + instructions

### LINKED ISSUE #328: Tag commit when pushing to production
I suggest we use tag the last commit on the `production` branch every time we push the docs into production.

In my previous teams, we would always tag the commit that the production deployment was based on for reference. It came handy on a few occasions, namely when we needed to rollback to the previous 
working state, or when we wanted to checkout a specific file from the last deployment.

WDYT?

tradej:
The script automatically creates and applies the correct tag:

* `YYYY-MM-DD` if there hasn't been a release that day.
* `YYYY-MM-DD_NUMBER` if there has. `NUMBER` starts at 2 and is automatically incremented.

Known issues:

* The detection of the latest tag of today (line 17) does not work well when there has been more releases than 9.

Resolves #328.

### DIFF
--- a/RELEASE.adoc
+++ b/RELEASE.adoc
@@ -1,10 +1,31 @@
 # Release Instructions
 
-After the catalog update has been made, update the following lines with the latest version:
+After the catalog update has been made, do the following:
 
+* Update the following lines with the latest version:
++
+--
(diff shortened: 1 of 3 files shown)


rhoads-zach:
merged: https://github.com/openshiftio/appdev-documentation/commit/79af7d909fe7f6c750f1c8a8013c0be2236ba2f0

PR #381, opened 2017-07-25
Fixed previewDocsServer on Fedora & SELinux

### LINKED ISSUE #378: previewDocsServer.sh doesn't work in Fedora with SELinux on
The `docker` invocations in `$REPO_HOME/scripts/previewDocsServer` do not work in Fedora if SELinux is on because it is blocking access to the user's directory. To solve this problem, Docker needs to be run in privileged mode by adding the `--privileged` flag.

Also, the script invocation needs to be done either using an account in the `docker` group (which is a security hole, and is discouraged), or as `root`. It should either be documented that the script should be run as `root` either using `sudo` or `su -c`, or the `docker` invocations should be prepended themselves.

Notes:
* If the whole script is run using `sudo`, the `$HOME` directory resolves to `/root`, which effectively means that all the Maven packages will be downloaded into root's `.m2` directory.
* If the `docker` invocations are prepended with `sudo`, it is not obvious to the user why root access is required.
* When using `pkexec` instead of sudo, each invocation of `docker` requires the user to enter the password separately, which results in inferior user experience.

tradej:
The usage and caveats are documented in the Contributor Guide.

Resolves #378.

### DIFF
--- a/docs/topics/contributing-starting-preview-server.adoc
+++ b/docs/topics/contributing-starting-preview-server.adoc
@@ -1,0 +1,34 @@
+
+= Starting Preview Server
+
+You can preview the Launchpad locally. To do that, perform the following steps:
+
+[discrete]
+== Prerequisites
+
+* Docker installed.
+* Docker daemon running.
+
+[discrete]
+== Procedure
+
+.Running Preview Server Locally
+. As a regular user, run the `$REPO_HOME/scripts/previewDocsServer.sh` script from the repository's base directory:
++
+--
+[source,options="nowrap"]
+----
+$ ./scripts/previewDocsServer.sh
+----
+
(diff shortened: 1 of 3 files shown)


docs/topics/contributing-starting-preview-server.adoc
```diff
@@ -0,0 +1,34 @@
+
+= Starting Preview Server
+
+You can preview the Launchpad locally. To do that, perform the following steps:
```
Ladicek: I think the Launchpad word is discouraged, and also this doesn't really preview the Launch app -- it previews the documentation.

tradej: Thanks, changed.

rhoads-zach:
merged: 877f5d2dd6abba74ea2228fac265b1e5d0055600

PR #382, opened 2017-07-25
issue #304 alternate CURL commands for secured booster interaction

### LINKED ISSUE #304: Add curl commands to SSO booster interaction
Per conversation with @Ladicek, add interaction examples using raw `curl` commands in addition to using the provided script.

@Ladicek please provide any additional detail if i missed anything.

inoxx03:
sorry, wrong number, was supposed to be 304 instead of 370
resolves #304

### DIFF
--- a/docs/topics/secured-mission-sb-interaction.adoc
+++ b/docs/topics/secured-mission-sb-interaction.adoc
@@ -9,7 +9,7 @@ secured-springboot-rest,secured-springboot-rest-myproject.192.168.99.100.nip.io,
 sso,sso-myproject.192.168.99.100.nip.io,,sso,<all>,
 |===
 
-The Secured booster endpoint name is also available using the Java client:
+Alternatively, obtain the Secured booster endpoint name using the Java client:
 
 [source,bash,options="nowrap",subs="attributes+"]
 ----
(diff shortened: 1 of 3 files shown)


docs/topics/secured-mission-sb-interaction.adoc
```diff
@@ -37,6 +45,14 @@ Requesting greeting...
   "content" : "Hello, World!"
 }
 ----
++
+Alternatively, call the `/greeting` endpoint directly using the following command:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+curl -k -v http://secured-springboot-rest-{project-name}-latest.192.168.42.161.nip.io/greeting -H "Authorization:Bearer $MY_TOKEN"
```
Ladicek: The IP address here is different than in the rest of the document.

docs/topics/secured-mission-sb-interaction.adoc
```diff
@@ -52,6 +68,14 @@ Requesting greeting...
   "content" : "Hello, Scott!"
 }
 ----
++
+Alternatively, call the `/greeting` endpoint directly using the following command:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+curl -kv http://secured-springbot-rest-vx-secured-demo-latest.192.168.42.161.nip.io/greeting?name=Scott -H "Authorization:Bearer $MY_TOKEN"
```
Ladicek: Same as above -- different IP address. Also should possibly use `{project-name}` like above?

inoxx03: Thanks, for catching this. I already updated the examples to use {my-project} instead of the project name. I'll make the IP addresses consistent in the next commit

Ladicek:
The other interactions (invalid password, another user) could have the `curl` examples as well, but I agree it should be obvious from previous sections.

docs/topics/secured-mission-sb-interaction.adoc
Ladicek: Right, it also applies to the other runtimes I think.

inoxx03:
Thanks for the feedback:) 
We can still add the remaining examples later, if anybody asks for them:)

rhoads-zach:
thanks @inoxx03 merging.

rhoads-zach:
Resolves #304 

merged 6f921fd036a2f123092b2c61f114a0971635c768

Issue #288, opened 2017-06-16
version number showing right under the main heading


Ladicek:
In #254, I added a current Git revision to the footer of each documentation page. I didn't notice, however, that it also gets added right under the main heading. I think it's OK to have it in the footer, but it looks strange on such prominent place as the main page heading. Is there a reasonble way to remove it from the heading and keep it in the footer? We probably don't want to maintain our own copy of Asciidoc HTML template...

PR #391, opened 2017-07-26
Release tagging script and instructions

### LINKED ISSUE #328: Tag commit when pushing to production

tradej:
**Please re-review**.

Changes:

* RELEASE.adoc updated with information about how to tag
* Added the tagRelease.sh script
* Jenkins includes RFC 3339 date instead of the commit hash

Main difference from the previous (reverted) solution: tagging is manual (the same script is used as before, run it on the `production` branch after master has been merged in) and Jenkins puts the current date instead of the latest tag into the `:revnumber:` attribute. I know it is not as bulletproof as intended, but I think it should do for now, or at least until we find out a better way to automate the process.

Resolves #328.

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -2,7 +2,7 @@ node("docsbuilder-maven") {
   checkout scm
   stage("Build Documentation") {
     sh "pwd && ls -l"
-    sh "echo :revnumber: \$(git rev-parse --short=10 HEAD) >> ./docs/topics/templates/document-attributes.adoc"
+    sh "echo :revnumber: \$(date --rfc-3339=date) >> ./docs/topics/templates/document-attributes.adoc"
     sh "echo :SegmentTrackerToken: \$LAUNCHPAD_TRACKER_SEGMENT_TOKEN >> ./docs/topics/templates/document-attributes.adoc"
     sh "./scripts/buildGuides.sh"
     sh "mkdir -p ci/openshiftio-appdev-docs/src/main/resources/webroot"
--- a/scripts/tagRelease.sh
+++ b/scripts/tagRelease.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Not tagging if HEAD is already tagged
+if git name-rev --tags --name-only --no-undefined HEAD &>/dev/null; then
+	read -p "Your latest commit is tagged. Do you really want to re-tag? [yN] " choice
+	case $choice in
+		[Yy] ) :;;
+		* ) exit;;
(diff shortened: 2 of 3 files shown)


ci/Jenkinsfile
```diff
@@ -2,7 +2,7 @@ node("docsbuilder-maven") {
   checkout scm
   stage("Build Documentation") {
     sh "pwd && ls -l"
-    sh "echo :revnumber: \$(git rev-parse --short=10 HEAD) >> ./docs/topics/templates/document-attributes.adoc"
+    sh "echo :revnumber: \$(date --rfc-3339=date) >> ./docs/topics/templates/document-attributes.adoc"
```
Ladicek: Couldn't this read the Git tag?

Ladicek: Aha, sorry, didn't read the PR description :-)

scripts/tagRelease.sh
```diff
@@ -0,0 +1,34 @@
...
+		* ) exit;;
+	esac
+fi
+
+# Parsing current tag
+
+# Using name-rev is not good because it only gives you one tag, and not
+# necessary the latest version
+today_tag="$(date --rfc-3339=date)"
+latest_today_tag="$(git tag | grep "$today_tag" | sort | tail -1)"
```
Ladicek: Note that this will (I think) fail with number 10, as that would sort before 2 :-) Not a problem in this case I think.

ci/Jenkinsfile
Ladicek: Or maybe we could keep the SHA1 there for now. Your call :-)

tradej: I think this is ultimately @rhoads-zach's call, but having the ISO date sorta solves #288, as you said yourself. I'm ambivalent towards this.

scripts/tagRelease.sh
tradej: It likely will. I don't expect more than 9 releases in the same day, TBH.

rhoads-zach:
merged: 412c7f087d148612a1607cc062c3f43462f4ef17

ci/Jenkinsfile
Ladicek: Yea agree about #288.

PR #393, opened 2017-07-27
Replaced date invocation with portable options

### LINKED ISSUE #392: Release tagging does not work on UNIX machines
Following up on #391:

As reported by @rhoads-zach in the documentation Slack channel, the `$REPO_HOME/scripts/tagRelease.sh` script does not work on machines with OpenBSD or MacOS because the `date` binary there does not have the `--rfc-3339` flag. I assume this problem also affects the `ci/Jenkinsfile` script as I have used the same invocation there.

I believe the solution is to use `date "+%Y-%m-%d"` instead. I have verified it works on OpenBSD and Fedora. @rhoads-zach or someone with an Apple PC, can you please verify it works on MacOS as well?

tradej:
Please verify this code work on all platforms we intend to use. I have tried Fedora and OpenBSD and they work.

Resolves #392.

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -2,7 +2,7 @@ node("docsbuilder-maven") {
   checkout scm
   stage("Build Documentation") {
     sh "pwd && ls -l"
-    sh "echo :revnumber: \$(date --rfc-3339=date) >> ./docs/topics/templates/document-attributes.adoc"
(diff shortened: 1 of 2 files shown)


rhoads-zach:
@tradej LGTM. Script works on macOS

rhoads-zach:
merged: 79187b71e169cd5614377119627ecd4dc398d8ae

PR #437, opened 2017-08-08
revnumber is now using {localdate} variable

gastaldi:


gastaldi:
Let's use the output of git describe instead

PR #473, opened 2017-08-11
Fixes #469: Analytics token is replaced at deployment time

### LINKED ISSUE #469: Assign analytics token at deployment time
The current way of setting the analytics token (see https://github.com/openshiftio/appdev-documentation/blob/master/cico_build_deploy.sh#L54) does not work on ci.centos.org because the staging and production containers are identical, and this substitution happens at build time. This means that we end up with the staging token in production.

A possible solution is to have a placeholder in the template, and set the container's entry point to a script that replaces this placeholder (e. g. using `sed`) with either the staging token or the production token in all HTML files, based on an environment variable. The variable is supplied when the container is deployed.

gastaldi:


Ladicek:
Though not sure why I was appointed :-)

scripts/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/bash
+
+sed -i "s#\${LAUNCHPAD_TRACKER_SEGMENT_TOKEN}#$LAUNCHPAD_TRACKER_SEGMENT_TOKEN#" /usr/share/nginx/html/docs/*.html
+
+exec /run.sh
```
vpavlin: Where is run.sh?

Ladicek: Inside registry.centos.org/kbsingh/openshift-nginx:latest

vpavlin: Ah, yes, obviously, missed the FROM:)

tradej:
@Ladicek I wanted a third set of eyes, thanks for the review.

tradej:
Seems like all discussion was resolved, so LGTM then. @inoxx03, can you please merge?

inoxx03:
merging

inoxx03:
merged 08f1e186a0e99b53d304dce19f8f4e8487a677ba

PR #476, opened 2017-08-14
Testing in CICO build script

### LINKED ISSUE #471: Make testing an integral part of the build process
Currently, the `scripts/buildGuides-CI.sh` test script is not run when the containers are built. Testing the books should be an integral part of the build process. Make changes to the `cico_build_deploy.sh` script so that it either contains or calls `scripts/buildGuides-CI.sh` and can be used in the http://ci.centos.org build chain. 

Partly resolves #297.

tradej:
Testing is moved from the `buildGuides-CI.sh` script to the `cico_build_deploy.sh` script before the build process, so if any book fails to validate, the process fails.

The `buildGuides.sh` script is deleted becauase it is no longer necessary.

Doesn't have to be included in Aug 14 release.

Resolves #471.

### DIFF
--- a/cico_build_deploy.sh
+++ b/cico_build_deploy.sh
@@ -28,6 +28,10 @@ function tag_push() {
 # Exit on error
 set -e
 
+# Run tests
+./scripts/validateGuides.sh
+
+
 if [ -z $CICO_LOCAL ]; then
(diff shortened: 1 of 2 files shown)


inoxx03:
merged: 952ab633ee83ec1c4fc2dcaef43ccea24c7ed3df

Issue #439, opened 2017-08-08
Remove unnecessary files


tradej:
Once the Nginx move is finished, the following files or directories will not be necessary any more, and can be purged:
```
ci/
scripts/dockerTagRelease.sh
```

Dependent on #428.

PR #478, opened 2017-08-14
Using date instead of Git to set version name

### LINKED ISSUE #470: An alternative way of setting :revnumber:
The current way of obtaining `:revnumber:` (i. e. the last Git tag in the repo) does not work on http://ci.centos.org because the `.git` directory is not copied to the machine that runs the `cico_build_deploy.sh` script that builds the doc set. There needs to be an alternative way of setting this variable. Off the top of my head, I can think of these:

* **Set the current date.** This has the downside that the date in the documentation may not correspond with the tag in the repository (when the container is rebuilt later than that day). However, I don't think this is a big issue because any time we need to revert to the previous stable release, we just go back in Git history and the first tag we come across will be the latest stable version.

* **Maintain versioning manually.** Meaning that the `:revnumber:` attribute would be hardcoded in `document-attributes.adoc`. The obvious downside is that this must be part of the release train, but my experience from the BxMS team shows that it is not much of a hassle.

* ~~**Set the commit hash.** This is rather ugly and user unfriendly, but also unambiguous.~~

All of these solutions still expect tagging to happen because we need to track which commits we published from in the past.

tradej:
NOTE: Replacing in `ci/Jenkinsfile` too in case anyone used it for anything (they shouldn't, though, and the file will be deleted in #439).

Resolves #470.

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -2,7 +2,7 @@ node("docsbuilder-maven") {
   checkout scm
   stage("Build Documentation") {
     sh "pwd && ls -l"
-    sh "echo :revnumber: \$(git describe --abbrev=0 --tags) >> ./docs/topics/templates/document-attributes.adoc"
+    sh "echo :revnumber: \$(date '+%Y-%m-%d') >> ./docs/topics/templates/document-attributes.adoc"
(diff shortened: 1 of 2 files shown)


rhoads-zach:
thanks for the quick turnaround @tradej. LGTM, merging.

tradej:
Welcome.

gastaldi:
I'd rather use {localdate} in the doc-attributes and remove that line, but that's fine :)

PR #487, opened 2017-08-15
Installing Asciidoctor in CI build

### LINKED ISSUE #486: Testing always fails in the CI build
https://ci.centos.org/job/devtools-appdev-documentation-generator-build-master/26/console

tradej:
The CI build environment is missing Asciidoctor. Adding the installation thereof.

I have tested this on Fedora (bare metal), CentOS (in Docker), and UNIX (FreeBSD, assuming `gem` is installed).

Resolves #486.

### DIFF
--- a/cico_build_deploy.sh
+++ b/cico_build_deploy.sh
@@ -28,6 +28,13 @@ function tag_push() {
 # Exit on error
 set -e
 
+# Install Asciidoctor if missing - needed for CI
+if ! asciidoctor --version &>/dev/null; then
+    echo Asciidoctor is missing, installing...
+    yum --version &>/dev/null && yum -y install rubygems
+    gem install asciidoctor
+fi
+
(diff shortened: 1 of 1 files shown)


gastaldi:
Isn't that supposed to happen inside the Dockerfile.build?

vpavlin:
@gastaldi That's a good point. Maybe running tests directly in Dockerfile.build during build would make things more transparent and portable. What do you thing @tradej ?

Other option is what @tradej does here - make sure asciidoc is installed and then run tests from the build script, but it might make sense to consider saving time and resources by not generating the html twice

inoxx03:
merged: b40e58cf0c31574dee64855901a7e403a2859bfe

tradej:
@gastaldi @vpavlin I have filed a follow-up issue for moving the testing part to the container.

PR #498, opened 2017-08-15
Resolves #495: Fixed dead anchors in links

### LINKED ISSUE #495: Update link targets in launcher review screen

rhoads-zach:


gastaldi:
According to @edewit, `document-attributes.adoc` is included in all transformations

rhoads-zach:
@tradej @gastaldi For some reason i thought that was still an issue. Sorry for the confusion. Updated based on this info.

docs/topics/circuit-breaker-mission-intro-paragraph.adoc
```diff
@@ -1,3 +1,6 @@
-Mission proficiency level: xref:proficiency_foundational[*{proficiency-foundational}*].
+Mission proficiency level: 
+//special case since topic is used by front end.
+ifdef::docs-topic[xref:proficiency_foundational[*{proficiency-foundational}*].]
+ifndef::docs-topic[link:https://appdev.openshift.io/docs/wf-swarm-runtime.html#proficiency_levels[*Foundational*].]
```
tradej: I'm afraid this still needs to be changed to `{proficiency-foundational}` (in all files).

In addition to that, the URL is hard-coded to point to WF Swarm runtime guide. Is that intentional?

edewit: @tradej it's defined here: https://github.com/openshiftio/appdev-documentation/blob/2abfe1aaa7f82fb2ed885bcc5e6c301ec28224f7/docs/topics/templates/document-attributes.adoc#L132 and that works
screenshot from 2017-08-17 13-26-32

edewit: @tradej @rhoads-zach so I need to set the `:docs-topic:` variable to what?

tradej: @edewit: No, you should not set `:docs-topic:` at all, setting this variable enables docs-only stuff. In the launchpad, you want it unset at all times.

Re foundational: In case of the launchpad line (L4), the string is hardcoded to `Foundational`. We only need it to use the same attribute as the docs line (L3), i. e. `{proficiency-foundational}`. No changes in the launchpad should be necessary as the value of this attribute is defined in `document-attributes.adoc`.

Re URL: If we want to link to the appropriate guide based on what runtime the user selected on the launchpad, we'll need you to set some attribute that'd help us generate the appropriate URL based on info in `document-attributes.adoc` (i. e. which runtime is being used). Having said that, if pointing to the WF Swarm book only is fine (the information in the appendix is identical in all guides after all), no changes need to be done.

rhoads-zach: Thanks @tradej, nice catch on the last var, sorry I missed that.

Re: swarm guide, yes, thats intentional.  I figured it was a sensible default since we can't know the runtime from the front end without some extra work and ifdef logic.

rhoads-zach: @edewit you shouldn't have to take any additional action. We built this so the topic itself could figure out if it was being used by a doc or the front end.

tradej:
LGTM now.

inoxx03:
merging

inoxx03:
merged: 2bc4a138f1968b7dc415fffc0e63313bbdab660b

PR #503, opened 2017-08-16
Moved testing into the build container

### LINKED ISSUE #490: Move testing into Dockerfile.build
Move testing the books from `cico_build_deploy.sh` to `Dockerfile.build` for better portability and deduplication of code (making sure Asciidoctor is installed).

Follows up on #487.

tradej:
Adding `libxml2` just in case it gets removed from the base CentOS image.

I have tested that if there is a change to the sources, rebuilding indeed triggers revalidation (i. e. the Docker image is not cached).

Resolves #490.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -5,9 +5,12 @@ MAINTAINER Zach Rhoads (https://github.com/rhoads-zach)
 USER root
 
 # Install asciidoctor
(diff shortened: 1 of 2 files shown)


inoxx03:
merging

inoxx03:
merged: dd6869c2e76907dbc06b1364d13ab69fdaedc7a4

PR #520, opened 2017-08-21
Fixes #517: added dir permission setting to script

### LINKED ISSUE #517: cico_build_deploy.sh creates the "html" directory as root
When the `cico_build_deploy.sh` script is run, it builds the doc set inside the container with root privileges, therefore creating the `$REPO_HOME/html` directory owned by `root:root`. This makes it impossible for the regular user to build the doc set later using the `build_guides.sh` script as it tries to remove the root-owned directory.

Either the `html` directory should be removed after it serves its purpose in the `cico_build_deploy.sh` script, or its ownership should be set to the regular user who ran the script.

rhoads-zach:
@tradej let me know what you think.

### DIFF
--- a/cico_build_deploy.sh
+++ b/cico_build_deploy.sh
@@ -52,9 +52,16 @@ rm -rf ${TARGET_DIR}/
 docker build -t ${BUILDER_IMAGE} -f Dockerfile.build .
 
 mkdir ${TARGET_DIR}/
+mkdir ${TARGET_DIR}/images
+
 docker run --detach=true --name ${BUILDER_CONT} -t -v $(pwd)/${TARGET_DIR}:/${TARGET_DIR}:Z ${BUILDER_IMAGE} /bin/tail -f /dev/null #FIXME
 
+
 docker exec ${BUILDER_CONT} sh scripts/build_guides.sh
+
+#Need to do this again to set permission of images and html files
+chmod -R 0777 ${TARGET_DIR}/
+
 
 #BUILD DEPLOY IMAGE
 docker build -t ${DEPLOY_IMAGE} -f Dockerfile.deploy .
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 html
 **/*.DS_Store
+**/master.xml
 .idea
 **/*.iml
 **/*.idea


rhoads-zach:
fixes #517

cico_build_deploy.sh
```diff
@@ -52,6 +52,11 @@ rm -rf ${TARGET_DIR}/
 docker build -t ${BUILDER_IMAGE} -f Dockerfile.build .
 
 mkdir ${TARGET_DIR}/
+mkdir ${TARGET_DIR}/images
+
+#This sets the permissions on target dir since this script may be invoked as root, but container is not.
+chmod -R 0777 ${TARGET_DIR}/
```
tradej: This invocation needs to be after line 57 (the `build_guides.sh` invocation) because the files created by the build are then still owned by root and therefore impossible to delete by the regular user.

rhoads-zach:
thanks for the feedback @tradej. Updated.

cico_build_deploy.sh
```diff
@@ -52,10 +52,17 @@ rm -rf ${TARGET_DIR}/
 docker build -t ${BUILDER_IMAGE} -f Dockerfile.build .
 
 mkdir ${TARGET_DIR}/
+mkdir ${TARGET_DIR}/images
+
+
 docker run --detach=true --name ${BUILDER_CONT} -t -v $(pwd)/${TARGET_DIR}:/${TARGET_DIR}:Z ${BUILDER_IMAGE} /bin/tail -f /dev/null #FIXME
 
+#This sets the permissions on target dir since this script may be invoked as root, but container is not.
+chmod -R 0777 ${TARGET_DIR}/
```
tradej: This needs to happen _after_ `docker exec ${BUILDER_CONT} sh scripts/build_guides.sh` because that script creates files in the `html` directory as root, so even if you run `chmod` now, the regular user won't be able to delete them later.

rhoads-zach:
thanks @tradej, Fixed and added gitignore as well

cico_build_deploy.sh
```diff
@@ -52,10 +52,20 @@ rm -rf ${TARGET_DIR}/
 docker build -t ${BUILDER_IMAGE} -f Dockerfile.build .
 
 mkdir ${TARGET_DIR}/
+mkdir ${TARGET_DIR}/images
+#This sets the permissions on target dir since this script may be invoked as root, but container is not.
+chmod -R 0777 ${TARGET_DIR}/
```
tradej: I know it may look like nitpicking, but it seems to me you forgot to delete this line as the same invocation is now (correctly) on L66. :)

rhoads-zach:
Reran and was unable to reproduce. Fixed and merging.

PR #530, opened 2017-08-24
Fix for validate_guides.sh on Mac OS

### LINKED ISSUE #529: validate_guides.sh does not work on a Mac
The different behaviour of `dirname` on a Mac as opposed to Linux or BSD makes the script succeed even when no tests were run.

The error is:
```
usage: dirname path
```
Likely caused by invoking dirname with multiple arguments that aren't put in quotation marks or some such.

tradej:
Resolves #529. I have included some cleanup so that it is immediately testable on a vanilla Mac.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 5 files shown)


scripts/entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
```
gastaldi: I think the most compatible option is: 
`#!/usr/bin/env bash`

tradej: Thanks, @gastaldi, for the tip. Fixed.

rhoads-zach:
merged: 274cad8654a66634cbb10b895b36fb514eefc6e4

PR #531, opened 2017-08-24
Added pre-commit hook to prevent breaking Launchpad

### LINKED ISSUE #505: Create a commit hook to prevent breaking launchpad-frontend
The [openshiftio/launchpad-frontend](https://github.com/openshiftio/launchpad-frontend) repository uses some of the files in this repository. Those files are listed in the [adoc.index](https://github.com/openshiftio/launchpad-frontend/blob/master/src/assets/adoc.index) file in the openshiftio/launchpad-frontend repository.

We should set up a commit hook that would warn the user in case they removed the files in question.

Considerations:
* The commit hook must access the file every time it is run, therefore we must ensure a fallback if the user is not online when committing.
* The file is a JSON file, but in the name of portability, I believe we should parse it with lowest-common-denominator command-line tools.

tradej:
This hook must be run client-side (GitHub does not allow server-side hooks, only webhooks to other services.). This means that since the hooks are in the `.git` directory and therefore not tracked by Git, they must be put in another directory (`.githooks` in this case) and the user must configure their repo instance accordingly. This is described in the Contribution Guide.

We can set up server-side checking in the `cico_build_deploy.sh` script, I believe, to verify all required files _exist,_ as opposed to _were not removed in the last commit,_ which is what the commit hook does. I realise that is a slightly different thing, but I think it's close enough.

Resolves #505.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+LAUNCHPAD_INDEX="https://raw.githubusercontent.com/openshiftio/launchpad-frontend/master/src/assets/adoc.index"
+DOC_REPO_PREFIX="https://raw.githubusercontent.com/openshiftio/appdev-documentation/master/"
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+	against=HEAD
+else
+	# Initial commit: diff against an empty tree object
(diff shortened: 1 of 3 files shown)


rhoads-zach:
LGTM, merging.

rhoads-zach:
merged: 80215023d334e4fb64e59ea1ee295ec4b145b78a

PR #537, opened 2017-08-25
Added script to generate changelog for a milestone

### LINKED ISSUE #456: Investigate ways to automate changelog generation and updates using the GitHub API.
https://developer.github.com/v3/issues/

tradej:
Notes:

## Gitignore
* The `.github_token` file is used to store the current user's GitHub access code for making queries. The file is optional—the user can provide the token interactively or not at all. GitHub's quota may come into play in the last scenario, though.
* `CHANGELOG.temp` is a Markdown temp file that the LogHub binary creates when querying, even if the output is directly piped somewhere.

## Contrib Guide
* I moved the bullet point with updating the milestone earlier in the process so that the issues resolved can be put into CHANGELOG and committed with other changes.

## The Script
* I have tested on Fedora, FreeBSD, and Mac OS. Having said that, I'll have to return the Mac soon, so I will go back to testing on FreeBSD only.
* The script only outputs the list of issues to `stdout`, and the result must be copied to `CHANGELOG.adoc` manually. Manipulating the file itself using the script has proven to be quite tricky in regard to corner cases and assumptions. I don't think it saves enough time to warrant the extra work.

Resolves #456.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@ Guardfile
 .settings
 *.orig
 *.swp
-
+.github_token
+**/CHANGELOG.temp
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,12 +4,13 @@
 #specific files
 docs/contrib-guide/master.adoc @tradej
 openshift/* @gastaldi
-**/*.buildGuide.sh @rhoads-zach
-scripts/buildGuides.sh @rhoads-zach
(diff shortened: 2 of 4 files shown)


rhoads-zach:
merging.

rhoads-zach:
merged: 36501f53e1c87f9d92f9b530897a7087a2efc7fa

PR #569, opened 2017-08-30
Issue #563 Fixed POST request description in CRUD mission intro.

### LINKED ISSUE #563: POST requests are not executed on /api/fruits/* endpoint
Spring Boot CRUD booster's documentation says that POST requests should be executed on `/api/fruits/*` by passing a valid ID. That is not correct because when creating a fruit user cannot know an ID which will be assigned to a newly created fruit. POST requests should be executed on `/api/fruits`.

inoxx03:
Section now states that the user should pass in a valid value for ```name``` instead of an ID when creating a new entry.
Also  removed the asterisk from  the endpoint  name.

Resolves #563 

@gytis Please let me know what you think.


tradej:
[test]

rhoads-zach:
merged: 16fea929dd2bdf8cb809a75af36a3fdfbf34f892

PR #571, opened 2017-08-30
fixes #565

### LINKED ISSUE #565: Is it OK to say that booster needs to be created with launch.openshift.io when it's not on OSIO?
Every Spring Boot booster documentation says that it first has to be created with `launch.openshift.io` before deploying it with `oc` command. However, not all of them are available on actual `launch.openshift.io` but on it's copy running locally. It might be a bit confusing for the user where to create a project.

rhoads-zach:


rhoads-zach:
@gytis i updated the warning message to be a bit more explicit about using the application. This covers both zip and ci/cd flow  as well as oso and minishift.

tradej:
[test]

gytis:
LGTM

rhoads-zach:
merged: 602df267b62dc84623b77928093c54f3abae97c7

PR #604, opened 2017-09-01
fixes #594

### LINKED ISSUE #594: Getting Started / 1. Create a Booster: maybe note that login is required
In the 2nd step of the launch.openshift.io wizard at latest, you have to login using Red Hat Developers account. Maybe we should call it out in this chapter.

rhoads-zach:


rhoads-zach:
merged: 278d62f7d0464a330560b046cae6abd2fb3d565d

PR #605, opened 2017-09-01
updated number of services started

### LINKED ISSUE #599: Minishift Guide / 4.2. Manually: one small issue
Near the end of this chapter, in item 10 of the procedure, the text reads "wait and confirm that the three services for the launch.openshift.io application are starting up". And below, there's a picture with _four_ services. (Which is correct, there _are_ 4 services.) Just change the number in that sentence :-)

lavernw1:
Fixes #599. updated number of services in step 10


rhoads-zach:
merged: 35a4a4fa5043c55b38ec1a3d3b8c85f2b9939680

PR #606, opened 2017-09-01
fixes #595

### LINKED ISSUE #595: issues in README of WildFly Swarm REST API Level 0 .zip booster
I see a couple of small issues in the generated README in the .zip of the Swarm REST booster. I guess some/most of them are more common than for just this booster.

1. It says "unzip booster-rest-http-wildfly-swarm.zip", but the .zip name is "whatever-artifact-id-i-chose-in-the-wizard.zip".
2. It says "cd booster-rest-http-wildfly-swarm" after unzipping, but the directory is called "whatever-artifact-id-i-chose-in-the-wizard".
3. In "Update this Booster", it says "Stop your booster" but doesn't tell how. (It's Ctrl+C on Linux, not sure about Windows.)
4. It says "mvn clean fabric8:deploy -Popenshift -DskipTests", but there's really no reason to skip the tests. They pass just fine. (They are just local tests, as we don't specify `-Popenshift-it` here.)
5. There's also a section "CI/CD Deployment", which is just wrong for the .zip. But the last sentence "You can also learn more about this booster and rest of the WildFly Swarm runtime in ..." makes sence and should be kept.

rhoads-zach:


inoxx03:
LGTM, merging

inoxx03:
merged c58aa5a4e09879599a9155e860365812f5750f9c

PR #610, opened 2017-09-01
Validating in pre-commit hook

### LINKED ISSUE #553: Validate books in the pre-commit hook
At the moment, we require contributors to manually run the book verification script before they make a contribution. With the pre-commit hook in place (see #505), we can automate this process to run every time the user commits a change.

Notes:
* The output of the validation script should be differentiated from the Git output, for example by prepending `[POLICY] ` or a similar string to every line of the output.
* The script must fail gracefully and allow the commit in the case when the failure of the validation script is not due to a book being malformed, but due to a fact that a binary is missing or the script is run on an unsupported architecture.

tradej:
Resolves #553.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,7 +1,8 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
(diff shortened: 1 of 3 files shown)


rhoads-zach:
LGTM, merging.

rhoads-zach:
merged: e0fc5ea5af45f528bcf8ebc672eadd4d181f51a4

PR #613, opened 2017-09-02
Proper means to reference multiple mvn profiles

rdruss:
According to http://maven.apache.org/guides/introduction/introduction-to-profiles.html multiple profiles are referenced as a comma separated list:

"This option takes an argument that is a comma-delimited list of profile-ids to use."


Ladicek:
I think both works, but I personally prefer `-Popenshift,openshift-it` as well :-)

rhoads-zach:
[test]

inoxx03:
merged: 7ea11a0deb23aec6fbaa1ae43b5c050a156167e6

PR #622, opened 2017-09-06
Issue 601: factual corrections to the Nexus install procedure

### LINKED ISSUE #601: Minishift Guide / Appendix A: Using Nexus: couple of issues
At the beginning of the chapter, in subchapter A.1, the text reads "You must use oc version 3.6.0 or later". I've just tested with Minishift 1.3.1 and its default version of OpenShift Origin, which is 1.5, and it all works just fine. I think we should just remove this sentence and also, from the code block above, the command `minishift config set openshift-version v3.6.0`. I don't know who asked to add the 3.6 version callout -- in my experience, it's wrong.

The instructions in subchapter A.2 manually pull a bunch of Docker images, including the one for Nexus. In my experience, this step is completely unnecessary. But it's not wrong either. So maybe mark these steps as optional.

Then, in step 10 of the instructions in subchapter A.3, we have "Go to the section titled Kubernetes Pod Template, click on Add Evironmental Variable", but that's very much wrong. There can be (and will be) _multiple_ pod templates. We have to tell user to find the one called "launchpad-maven", as that's what boosters use.

Also in step 11, the value is of the env variable is specified as `nexus-NEXUS_PROJECT_NAME.LOCAL_OPENSHIFT_HOSTNAME/nexus/content/groups/public/`, but that's wrong as well, it needs to be a full URL. So just prepend `http://` to it. It must read `http://nexus-NEXUS_PROJECT_NAME.LOCAL_OPENSHIFT_HOSTNAME/nexus/content/groups/public/`.

inoxx03:
resolves #601

### DIFF
--- a/docs/topics/minishift-install-nexus-setup.adoc
+++ b/docs/topics/minishift-install-nexus-setup.adoc
@@ -7,45 +7,30 @@ While developing your cloud-native applications with Java and Maven you may be r
 You can deploy a Nexus Repository server alongside the {launcher} application on your {OpenShiftLocal} and use it to fetch artifacts from the Maven Central Repository and cache them locally.
 This helps you speed up your builds and rolling updates and alleviates network load during build time.
 
-== Prerequisites for Deploying Nexus on {OpenShiftLocal}.
+== Setting up Your {OpenShiftLocal} for using Nexus.
 
-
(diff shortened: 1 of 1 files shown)


rhoads-zach:
LGTM

rhoads-zach:
merged: 6b7914e163a9031cc6d5a2b76e543a25292a94ea

PR #624, opened 2017-09-06
Issue #616: added note about code download

### LINKED ISSUE #616: SpringBoot, VertX, Swarm booster chapters - note about code download
There should be some note at the beginning of every mission chapter where to download code what is chapter about.
For example 'Code can be downloaded from launchpad...select booster xy and download the code in zip....

Just to know and see code what is chapter about.

inoxx03:
resolves #616 

@liborfuka @Ladicek: Would a note like this be OK? Since we have the ZIP download instructions already documented I included a link to the appropriate section of the GSG.
Please let me know WDYT and if it's ok, I'll include the note at the top of each Mission chapter. 

As I mentioned in the discussion on #616, there is also the alternative of linking directly to the repo on GitHub that hosts the booster code.

### DIFF
--- a/docs/nodejs-runtime/master.adoc
+++ b/docs/nodejs-runtime/master.adoc
@@ -23,6 +23,8 @@ include::topics/dev-guide-mission-intro.adoc[leveloffset=+1]
 
 include::topics/rest-level-0-mission-intro.adoc[leveloffset=+3]
 
+include::topics/note-preview-booster-source.adoc[leveloffset=+3]
+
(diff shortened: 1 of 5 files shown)


Ladicek:
Yea I think this makes sense. No need to repeat the instructions we already describe elsewhere.

Ladicek:
LGTM. @liborfuka WDYT?

liborfuka:
I think its sufficient note and #616 can be resolved.

rhoads-zach:
merged: 08f03b521b529e4d700e8789af31598072ce0198

PR #626, opened 2017-09-07
Issue #577: add info about the location of the command line tools help section in the web console

### LINKED ISSUE #577: Is "Command Line Tools" section in an obvious location to new OpenShift users?
"You can find this in the Command Line Tools section your Single-node OpenShift Cluster Web console."

This implies that the user is aware that the section in question can be found in the question mark drop down to the left of their username.  If I was a new user, this location would be unintuitive to me and I may look to the left menu to find this section.

inoxx03:
Resolves #577 

@rdruss WDYT?

### DIFF
--- a/docs/topics/circuit-breaker-mission-deploy-booster.adoc
+++ b/docs/topics/circuit-breaker-mission-deploy-booster.adoc
@@ -8,14 +8,7 @@ include::booster-deploy-openshift-local.adoc[leveloffset=+1]
 
 include::oc-client-deploy-booster-download-note.adoc[leveloffset=+3]
 
-. Get your authentication token for using the `oc` CLI client with your {OpenShiftLocal} account. You can find this in the _Command Line Tools_ section your {OpenShiftLocal} Web console.
-
-. Authenticate your `oc` CLI client with your {OpenShiftLocal} account by using your authentication token.
-+
-[source,bash,options="nowrap",subs="attributes+"]
-----
-$ oc login {link-oso-auth} --token=MYTOKEN
-----
+include::booster-deploy-get-cli-token.adoc[levelloffset=+3]
 
 . Create a new project in your {OpenShiftLocal}.
 +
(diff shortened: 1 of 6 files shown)


docs/topics/circuit-breaker-mission-deploy-booster.adoc
```diff
@@ -8,9 +8,12 @@ include::booster-deploy-openshift-local.adoc[leveloffset=+1]
 
 include::oc-client-deploy-booster-download-note.adoc[leveloffset=+3]
 
-. Get your authentication token for using the `oc` CLI client with your {OpenShiftLocal} account. You can find this in the _Command Line Tools_ section your {OpenShiftLocal} Web console.
+. Get the command containing your authentication token for using the `oc` CLI client with your {OpenShiftLocal} account:
+.. Click on the question mark icon in the top right-hand corner of the Web console, next to your user name.
+.. Select _Command Line Tools_ in the drop-down menu.
+.. Click the button next to the first text box to copy the command containing the token.
```
Ladicek: Does the text box have a label or something? "First text box" isn't very specific.

Ladicek:
Not my call to make, but there's quite some duplication, maybe make it a shared topic? :-)

docs/topics/circuit-breaker-mission-deploy-booster.adoc
inoxx03: @Ladicek unfortunately, the text box doesn't appear to have a usable label. There is a text headline, but it is more of an instruction than a label. (image below for reference)
token 

I see two possible ways of working around this:
1.  We can provide an image for reference.
Admittedly rather clunky, because we have to update it if the UI changes.

2. We can provide a non-ambiguous descriptive identifier. 
Something like:

> The text box containing the login command with the hidden token.  


WDYT?

inoxx03:
Thanks. I agree, making it a shared topic is definitely a good idea:)

docs/topics/circuit-breaker-mission-deploy-booster.adoc
Ladicek: Yea agree, "text box containing the `oc login ...` command with a hidden token" sounds good.

inoxx03:
@rhoads-zach I put the procedure of obtaining the command with the token and logging in  with `oc` into a separate topic. WDYT?

rhoads-zach:
LGTM, merging.

PR #654, opened 2017-09-20
Do not import env as it's not used anyway

vpavlin:
This fixes failing builds for https://github.com/openshiftio/appdev-documentation/pull/609 - the problem there was env var `ghprbPullLongDescription` got imported (see https://ci.centos.org/job/devtools-appdev-documentation/87/console) and for some reason sourcing failed. 

I removed the env import as none of the env vars are used anyway.


tradej:
merged: a378f5f

PR #662, opened 2017-09-20
Storing XML schema locally for improved performance

### LINKED ISSUE #655: Improve validating performance on slow connections
The `validate_guides.sh` script now always downloads the DocBook schema when validating a single guide. This incurs a significant performance penalty on slow connections (11 seconds for the entire validation _on my 100 Mbps cable connection_ at home), which is a problem since validation is now run with every commit.

I believe downloading the schema once and using the stored file for validating every book should alleviate this problem significantly.

tradej:
Since the DocBook 5 schema is very unlikely to change, we can afford to store it locally. The performance differences are as follows:

When downloading:
* In the office: 1.2 seconds
* At home: 11 seconds

When stored locally:
* In the office: 1.2 seconds
* At home: 1.1 seconds

Disclaimer: My computer in the office is a different one from that at home.

Even if we need to change the schema for Pantheon to work, we can simply update and commit the XSDs in `$REPO_HOME/scripts/xml-schema` and keep on rolling.

Resolves #655.

### DIFF
--- a/scripts/validate_guides.sh
+++ b/scripts/validate_guides.sh
@@ -1,10 +1,15 @@
 #!/usr/bin/env bash
 
 # Builds all books into DocBook 5 XML and validates them using XMLlint.
+
+SCRIPT_SRC="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
+XML_SCHEMA="$SCRIPT_SRC/xml-schema/docbook.xsd"
+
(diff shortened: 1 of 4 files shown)


tradej:
@rhoads-zach, please test on Mac, thank you.

rhoads-zach:
merged: 284af5f99de0099da7410cd6384b11469c335cda

PR #664, opened 2017-09-21
Issue #641 updated nexus maven mirror steps for s2i builds

### LINKED ISSUE #641: Update nexus repository section based on S2I updates
https://appdev.prod-preview.openshift.io/docs/minishift-installation.html#_using_a_nexus_repository_server_with_the_launch_openshift_io_application_on_your_single_node_openshift_cluster

inoxx03:
fixes #641 
(branch name says 648, sorry, my bad:))

@cmoulliard please let me know what you think.

### DIFF
--- a/docs/topics/minishift-install-nexus-setup.adoc
+++ b/docs/topics/minishift-install-nexus-setup.adoc
@@ -1,5 +1,8 @@
 // name variable defined locally, because it is only used in this topic
 :nexus-project-name: NEXUS_PROJECT_NAME
+// e.g. https://github.com/path/to/source.git
+:scm-repo-url: SCM_REPOSITORY_URL
+
 
 = Using a Nexus Repository Server With the {launcher} Application on a {OpenShiftLocal}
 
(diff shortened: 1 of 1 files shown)


docs/topics/minishift-install-nexus-setup.adoc
```diff
@@ -121,55 +124,28 @@ Nexus comes pre-configured for the Maven Central Repository, but you may need ot
...
+Ensure that the YAML template of the builder image you are using for your application has the `MAVEN_MIRROR_URL` environmental variable defined.
+If it does not, see the Nexus documentation for instructions on link:https://help.sonatype.com/display/NXRM2/Maven+and+Other+Build+Tools[configuring your build manually^] before proceeding.
 --
 +
 [source,bash,subs="attributes+"]
 --
-NAME      HOST/PORT                                         PATH      SERVICES   PORT      TERMINATION
-jenkins   jenkins-{app-name}-{project-name}.{osl-route-hostname}             jenkins    <all>     edge/Redirect
+$ oc new-build {app-name}:latest~{scm-repo-url} \
+-e MAVEN_MIRROR_URL='http://nexus.{nexus-project-name}:8081/nexus/content/groups/public'
```
Ladicek: I wonder why `8081` now, it wasn't required before?

inoxx03: Not sure whether this is completely correct, but I based this procedure on a [similar procedure](https://docs.openshift.com/container-platform/3.5/dev_guide/app_tutorials/maven_tutorial.html) described in the Dev Guide for OCP 3.5. That procedure seems to require the port number, too.

Ladicek: Oh, I see it now -- this is a URL of the service, not of the route. (As in, a cluster-internal URL, which isn't reachable from outside. Which is OK in this case.) And indeed the service exposes port `8081`.

docs/topics/minishift-install-nexus-setup.adoc
```diff
@@ -121,55 +124,28 @@ Nexus comes pre-configured for the Maven Central Repository, but you may need ot
...
-The actual value of the `MAVEN_MIRROR_URL` environmental variable corresponds to the service route of the pod running the Nexus server instance on your {OpenShiftLocal}.
+. To ensure that your new build is using Nexus to retrieve artifacts, you can:
 
-To obtain this route, navigate to the project that contains your Nexus instance and execute the command `oc get routes nexus`.
---
-+
-. Click _Apply_ to confirm the configuration changes.
-+
-. Redeploy the Jenkins server instance with the latest configuration changes.
+* Navigate to `http://nexus-{nexus-project-name}.{osl-route-hostname}:8081/nexus/content/groups/public` to view the list of artifacts stored in your local repository.
```
Ladicek: But this won't work actually. I think. The Nexus _service_ exposes port `8081` (and that's what is used inside the cluser), but the Nexus _route_ (which is the only thing reachable from outside) exposes port `80` as usual. Do you have this rendered somewhere, so I can be sure I read this correctly?

inoxx03: Unfortunately, I not have a link to a rendered preview, but I've built the guide locally and uploaded the [HTML File](file.brq.redhat.com/home/brq/ssitani/public_html/minshift-installation.html)

Ladicek: OK, thanks. I'll write a more detailed comment.

Ladicek:
Looking at http://file.brq.redhat.com/~ssitani/minishift-installation.html#_setting_up_the_nexus_application, I see a couple of issues:

1. Item 5, "Find the external address of the Nexus instance", is no longer needed, you don't use it in the next step. Note that the notice "Nexus comes pre-configured for the Maven Central Repository, but you may need other repositories for your application" is useful and should be moved elsewhere.
2. Item 7 says "Pass in a parameter to set the value of the MAVEN_MIRROR_URL to match the route of your Nexus application", but what you're actually doing is passing the _service URL_, not the _route URL_.
3. Item 8 says "If your build uses Nexus to retrieve artifacts, the build log output should reference the path http://nexus:8081", but it actually won't be `http://nexus:8081`. That would only happen if Nexus was installed in the current project. The logs will show `http://nexus.NEXUS_PROJECT_NAME:8081/`.

TBH, I'm thinking maybe we should just link to https://docs.openshift.com/container-platform/3.5/dev_guide/app_tutorials/maven_tutorial.html, which is basically the same text.

inoxx03:
@Ladicek Thank you for the feedback:)
Personally, I also agree with your last point, that now that we no longer use a pipeline, the difference between this guide and  https://docs.openshift.com/container-platform/3.5/dev_guide/app_tutorials/maven_tutorial.html is minimal.  
We might as well just scrap this section altogether and link to the existing guide.

inoxx03:
@Ladicek I updated the procedure as per your feedback and moved the note from Step 5 to Step 7. Please let me now what you think.

docs/topics/minishift-install-nexus-setup.adoc
```diff
@@ -96,80 +99,40 @@ oc expose svc/nexus
...
-. Define a new environment variable:
-* Name: `MAVEN_MIRROR_URL`
-* Value: `nexus-{nexus-project-name}.{osl-route-hostname}/nexus/content/groups/public/`
-+
 [NOTE]
 --
-The actual value of the `MAVEN_MIRROR_URL` environmental variable corresponds to the service route of the pod running the Nexus server instance on your {OpenShiftLocal}.
-
-To obtain this route, navigate to the project that contains your Nexus instance and execute the command `oc get routes nexus`.
+Nexus comes pre-configured for the Maven Central Repository, but you may need other repositories for your application. To access images provided by Red Hat, add the `jboss-ga` Maven repository to your Nexus instance.
```
Ladicek: Can we make this a link, like in the OpenShift docs? The "jboss-ga" term isn't a widely known designator, I can see how some people might actually use `repository.jboss.org` based on this (which would be wrong). Actually instead of "jboss-ga", I'd rather say "Red Hat JBoss Middleware Maven Repository", which seems to be the official name.

inoxx03:
@Ladicek  I've added a more descriptive link title and included a link to the repository. also, removed the redundant port number from the route. WDYT?

docs/topics/minishift-install-nexus-setup.adoc
```diff
@@ -96,80 +99,40 @@ oc expose svc/nexus
...
-. Define a new environment variable:
-* Name: `MAVEN_MIRROR_URL`
-* Value: `nexus-{nexus-project-name}.{osl-route-hostname}/nexus/content/groups/public/`
-+
 [NOTE]
 --
-The actual value of the `MAVEN_MIRROR_URL` environmental variable corresponds to the service route of the pod running the Nexus server instance on your {OpenShiftLocal}.
-
-To obtain this route, navigate to the project that contains your Nexus instance and execute the command `oc get routes nexus`.
+Nexus comes pre-configured for the Maven Central Repository, but you may need other repositories for your application. To access images provided by Red Hat, add the link:https://maven.repository.redhat.com/ga/[Red Hat JBoss Middleware Maven Repository] to your Nexus instance.
```
Ladicek: Actually I'm thinking the Beta artifacts are in m.r.r.c/earlyaccess, not m.r.r.c/ga. Maybe the link should point just to https://maven.repository.redhat.com/ which has a description.

Ladicek:
Otherwise LGTM.

docs/topics/minishift-install-nexus-setup.adoc
inoxx03: @Ladicek Good point. I'll go with the link to https://maven.repository.redhat.com/. Thanks:)

cmoulliard:
Remark: Assigning 1Gb of disk space to the PVC definition for the Nexus Server is certainly not enough. It should be at least of 10Gb

inoxx03:
@cmoulliard Thank you for catching that:) I have updated the steps and command examples to specify a 10GiB PVC instead of 1 GiB.

tradej:
Is this good to merge?

inoxx03:
@tradej I'd say it is. @Ladicek approved the changes and I also fixed up the persistent volume claim size issue reported by Charles.

rhoads-zach:
@inoxx03 @tradej please hold this for now.

inoxx03:
@rhoads-zach is it OK if I un-WIP this?

tradej:
I'll merge this the first thing tomorrow.

tradej:
Merging...

tradej:
merged: a7bb378

PR #669, opened 2017-09-26
Glossary

### LINKED ISSUE #309: Create a Glossary
There may be developers reading our docs who are not familiar with some terms that we use, such as _opinionated framework, continuous delivery pipeline,_ or _reactive systems._

We should create a glossary that would explain such terms based on the expected knowledge level of the target audience. 

This glossary should be shared across the whole doc set.

*Possible sub-task: Get better description of the target audience from the stakeholders.*

tradej:
A short glossary with basic term explanations, all in one place. The explanations are the best I could come up with given my limited knowledge of the product, so please feel free to go to town on them.

The intention is to include the glossary at the end of all books for quick reference.

I have decided against implementing a glossary of terms unrelated to the product just yet, and try to reword the sections those terms if possible instead (see #670). We can discuss this with Ali in regard to long-term content strategy.

Resolves #309.

### DIFF
--- a/docs/topics/appendix-glossary.adoc
+++ b/docs/topics/appendix-glossary.adoc
@@ -1,0 +1,29 @@
+
+[[glossary]]
+= Glossary
+
+== Product and Project Names
+
+{launcher}:: The web application you launch boosters with.
+
+{OpenShiftLocal}:: An OpenShift cluster running on your machine using Minishift.
+
+== Terms Specific to {launcher}
+
+[glossary_booster]
+Booster:: An language-specific implementation of a particular xref:glossary_mission[mission] on a particular xref:glossary_runtime[runtime]. Boosters are listed in a xref:glossary_booster_catalog[booster catalog].
++
+For example, a booster is a web service with a REST API implemented using the {WildFlySwarm} runtime.
+
+[[glossary_booster_catalog]]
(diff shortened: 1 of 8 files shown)


docs/topics/appendix-glossary.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+== Terms Specific to {launcher}
+
+[glossary_booster]
+Booster:: An language-specific implementation of a particular xref:glossary_mission[mission] on a particular xref:glossary_runtime[runtime]. Boosters are listed in a xref:glossary_booster_catalog[booster catalog].
++
+For example, a booster is a web service with a REST API implemented using the {WildFlySwarm} runtime.
+
+[[glossary_booster_catalog]]
+Booster Catalog:: An Git repository that contains information about boosters.
```
Ladicek: Is that "a Git repository" instead of "an"?

Ladicek:
Otherwise LGTM.

docs/topics/appendix-glossary.adoc
tradej: Ha, good catch, thanks.

rhoads-zach:
merged: 38beeb6541948a49a2ea52a93b291d5a087563ca

PR #688, opened 2017-10-02
Issue #672: Update the title of the build process description

### LINKED ISSUE #672: "Building and Deploying to OpenShift Build Process" sounds weird
Appendix B of the Getting Started Guide is called "Building and Deploying to OpenShift Build Process", which sounds weird. This is after the recent heading changes; sorry for not catching it earlier.

inoxx03:
resolves  #672 

I realize that the new title is not very descriptive, but IMHO it could still work and doesn't sound as clunky as the previous one. Thoughts?

### DIFF
--- a/docs/topics/build-and-deploy-process.adoc
+++ b/docs/topics/build-and-deploy-process.adoc
@@ -1,6 +1,7 @@
 [[build-and-deploy-process]]
-= Building and Deploying to OpenShift Build Process
+= Building and Deploying to OpenShift
 
-When using {launcher}, you have the option to create and deploy a booster to OpenShift using the _Build and Deploy to OpenShift_ build process, which is based on link:{link-wf-swarm-runtime-guide}#s2i-build-process[the Source-to-Image (S2I) build process]. _Build and Deploy to OpenShift_ configures your booster's code to be built by OpenShift and deployed to OpenShift on each push to the master branch of your booster's GitHub repository. The benefit of this process is that it handles all the configuration, building, and deployment steps needed to get your booster's code running in OpenShift. It also allows you to quickly deploy code updates to help you rapidly iterate and see your changes in OpenShift. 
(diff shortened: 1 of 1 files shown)


Ladicek:
So the recent change, the one that caused this problem, updated all headings from infinitive, "Build and Deploy", to present participle, "Building and Deploying". (Sorry, I don't know the correct lingo here.) This PR moves one particular heading back, which is probably not what we want.

inoxx03:
@Ladicek Good point. Changing a single heading does indeed break up title consistency a bit. (Also, seems to me you do know the lingo :))

To elaborate on why I opted for this particular change in this title: 
If "Build and Deploy to OpenShift" is the name by which we refer to the CI/CD process used with RHOAR, this title could work. (If it is not the name we should use for our CI/CD process, then, of course, I'll change it immediately.)
(Same way as if the section were titled, say,  "Source-to-Image Build", we probably would not change it to "Buiding Images from Source" purely for the sake of consistency).

IMHO, using the present participle implies that the section contains instructions on completing a procedure.
This section, however, describes a process, rather than give instruction, so my aim was to make it less ambiguous with the title change.

Discalimer: I'm trying to avoid overly long titles, such as: "Build Process Description for Building and Deploying to OpenShift"

On the other hand, we could just go with "Building and Deploying to OpenShift". Not my first choice, but certainly brief and clear enough :)

tradej:
We definitely want to use gerunds in titles everywhere. For one, the IBM Style Guide mandates this, and it is the accepted way of doing this across CCS. This is also important because once modules are re-used across products, a deviation from the standard causes weirdness.

I request the heading be in the gerund form as long as there is a verb in it. _Building and Deploying to OpenShift_ works for me, even though it's technically a syllepsis (you don't _build to OpenShift_).

docs/topics/build-and-deploy-process.adoc
```diff
@@ -1,6 +1,7 @@
 [[build-and-deploy-process]]
-= Building and Deploying to OpenShift Build Process
+= Build and Deploy to OpenShift
```
tradej: Please keep the gerund form.

inoxx03: I am perfectly happy with that reasoning, Plus, it avoids the awful long title. Fixing...

inoxx03:
@tradej Syllepsis... nice catch:) I knew this figure was called a zeugma... turns out they are the same thing. Anyway, title corrected:)

Ladicek:
Haha, yea, "gerund" is the word I was looking for! Instead, I found "present participle". I love syllepsis, though! :-)

rhoads-zach:
@inoxx03 I think this will get clobbered due to the changes here:
https://github.com/fabric8-launch/appdev-documentation/pull/692/files

If there's anything specific you'd like me to incorporate, let me know. Otherwise, do you mind closing this?

inoxx03:
@rhoads-zach Thank you. It was only  a minor update due to a clunky section title. Seeing that you've already fixed the title in PR #692, I believe that this PR is outdated, so I'm closing it without merging.

PR #713, opened 2017-10-18
Script to sync WildFly Swarm sources

tradej:
The script does the following:

* Shallow-clones the WF Swarm repository in a termporary directory.
* Builds the upstream docs because some files we need are generated.
* Copies over files specified in the `scripts/wildfly-swarm-files.txt` to `docs/topics/wildfly-swarm` while keeping the original directory structure.
* Commits the files in the `docs/topics/wildfly-swarm` directory.
* Cleans up the temporary directory.

### DIFF
--- a/docs/topics/contributing-syncing-with-wildfly-swarm-docs.adoc
+++ b/docs/topics/contributing-syncing-with-wildfly-swarm-docs.adoc
@@ -1,0 +1,27 @@
+
+[#syncing-with-wildfly-swarm-docs]
+= Syncing with {WildFlySwarm} Docs
+
+Some documentation files for the {WildFlySwarm} runtim are sourced directly from the {WildFlySwarm} repository.
+You must synchronize them manually before a release to ensure the newest version of the files is available.
+
+The synchronization script deletes the existing `$REPO_HOME/docs/topics/wildfly-swarm` directory and replaces it with the latest version of the `master` branch from the upstream repository.
+Edit the variables in the `$REPO_HOME/scripts/sync_with_wildfly_swarm.sh` script to customize it.
+The variables are documented in the script.
+
+NOTE: Some files are not present in the upstream repository because they are generated. The script automatically builds the upstream project with Maven, ensuring these files are not missing.
+
+WARNING: Do not edit the files in the `$REPO_HOME/docs/topics/wildfly-swarm` directory, always submit a pull request to the upstream repository and synchronize the files when they are updated upstream.
+
+.Prerequisites
+
+* Maven installed
+
+.Procedure
+
+. Execute the `$REPO_HOME/scripts/sync_with_wildfly_swarm.sh` script:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ ./scripts/sync_with_wildfly_swarm.sh
+----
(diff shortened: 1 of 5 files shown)


docs/topics/contributing-syncing-with-wildfly-swarm-docs.adoc
```diff
@@ -0,0 +1,27 @@
+
+[#syncing-with-wildfly-swarm-docs]
+= Syncing with {WildFlySwarm} Docs
+
+Some documentation files for the {WildFlySwarm} runtim are sourced directly from the {WildFlySwarm} repository.
```
Ladicek: "runtim" -> "runtime"

scripts/sync_with_wildfly_swarm.sh
```diff
@@ -0,0 +1,47 @@
...
+git rm $FILES_DESTINATION
+
+echo "Cloning the WildFly Swarm repository..." >&2
+
+set -e
+
+pushd $TEMP_DIR &>/dev/null
+git clone $REPO_URL --depth 1 --branch $REPO_BRANCH
+cd $REPO_NAME/docs
+mvn clean install -DskipTests -Dswarm.product.build
```
Ladicek: Do we possibly want to have everything in community docs and `-Dswarm.product.build` in product docs? I'm fine with community docs documenting Swarm product, but it might be misleading (e.g. `bom-certified` doesn't exist in community Swarm).

tradej: 1) `swarm.product.build` is just a property, it's not a directory.
2) The upstream AsciiDoc sources are conditional to make sure we display the right content. See https://raw.githubusercontent.com/wildfly-swarm/wildfly-swarm/master/docs/howto/use-a-bom/index.adoc (look for `ifdef` and `ifndef`).

docs/topics/contributing-syncing-with-wildfly-swarm-docs.adoc
tradej: Thanks for catching that, I'll fix it in a follow-up pull request.

scripts/sync_with_wildfly_swarm.sh
Ladicek: Yea I know that Swarm can produce both upstream docs and product docs. The question is -- what version of Swarm docs do we include? Is it always product docs? (I'd guess yes.) Is that correct? (I'm not sure really.)

tradej: The only docs where this matters are the fractions reference docs, and those are included only from the product build. All other docs are included without modification, i. e. unbuilt.

Ladicek: That's not true, the explanation of BOMs is different between community Swarm and product Swarm too.

tradej: As I said before, please refer to https://raw.githubusercontent.com/wildfly-swarm/wildfly-swarm/master/docs/howto/use-a-bom/index.adoc. Both the community descriptions and the product descriptions are already contained in the file, therefore we need not make any further modifications. Displaying only the product-relevant information is done through setting the `:product:` attribute in the master file at build time.

Ladicek: OK, I still don't quite follow, but you know what you're doing, so I'll leave it at that and will file an issue later, if needed :-)

tradej: Let's meet tomorrow if you have time, I'll explain. This is an important mechanic in maintaining and building the docs, so I'd like everybody involved to know exactly what's up. I'll send you an invite tomorrow.

Ladicek: That makes sense. Thanks!

PR #723, opened 2017-10-20
Fixed output of validation script

tradej:
When a validation fails, the script outputs the full path of the book. With this fix, it outputs only the basename.


rhoads-zach:
LGTM, merging.

rhoads-zach:
merged: a4c468e84a325c9153a899c0b67f456dc00f7ee9

PR #724, opened 2017-10-20
Reintroduced scripts/preview_server.sh

### LINKED ISSUE #685: More user-friendly local preview server
Currently, the user must run two fairly complicated commands that do not take any arguments to get the local server running. I think we could reintroduce `scripts/preview_server.sh` that would do just that.

tradej:
Resolves #685.

The script is smart about permissions—if you run it as root, it does not require a password. If not, it runs `sudo`.


rhoads-zach:
merged: ccd304e79a11967699504347c8841583331dc1da

PR #727, opened 2017-10-23
WildFly Swarm upstream content (ENTSWM-181)

tradej:
New content for WildFly Swarm re using the runtime (see the `docs/wf-swarm-runtime/master.adoc` file).

Notes:

* I recommend reviewing the individual commits rather than the entire diff at once.
* I have gone with putting the information about the runtime usage under two first-level headings. I think it makes sense because this sort of information seems to have similar weight as the mission info, but if you want it anywhere else (i. e. second-level heading under Runtime Details, or anywhere else, really), just say the word.
* Likewise with Appendix C (fractions reference): I understand that the long list is unwieldy, but I see value in the user being able to access the individual fraction's section directly. If we value clean ToC more, I can fix it.
* The `:product:` attribute is set in the `docs/wf-swarm-runtime/master.adoc` file because the included WFS sources have a few conditional sections depending on whether you are building a community or a product build. This means that with this PR as is, only the product version is built. If we need to make this distinction in our builds, it's a matter of undefining it and/or moving it to `document-attributes.adoc` along with other attributes. I will file a follow-up issue so that we decide this before release.
* Likewise, there are two separate files with lists of the fractions in the `docs/wildfly-swarm/docs/reference` directory,  `index-community.adoc` and `index-product.adoc`, which include different fractions based on whether it's a community or a product build.
* I have disabled validation for the WFS guide because the DocBook 5 DTD does not recognize `&ndash;`, an entity used in the upstream docs . The usage of the entity is not a blocker for product builds; I have tried it. For more info, see [RHBZ#1097495](https://bugzilla.redhat.com/show_bug.cgi?id=1097495). I will file a follow-up issue to replace the usage of this element with something that generates valid DocBook XML.
* This issue resolves [ENTSWM-181](https://issues.jboss.org/browse/ENTSWM-181).


rhoads-zach:
merging.

rhoads-zach:
merged: 25f84554f62ef749cbd5743104ab2e90eb90538c

PR #750, opened 2017-10-31
Mod docs refactor

rhoads-zach:
@tradej @inoxx03 leaving this WIP until #743 and #744 are merged.  I would look at this by commit for a review perspective.

@Ladicek FYI, this has been a refactor with some very light editing.

PR #758, opened 2017-11-02
integration test steps for the Secured Spring Boot mission

inoxx03:
NOTE: This might need to be rebased after  #750 has been merged.

### DIFF
--- a/docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
+++ b/docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
@@ -1,10 +1,16 @@
+// This is a parameterized module. Parameters used:
+//
+//   parameter-runtime: runtime name in the module ID
+//
+// Rationale: This parameter displays the correct runtime name (`sb` or `vx`) in the module ID that is visible when linking to this section.
 
-[#running-the-secured-booster-integration-tests-on-vertx]
-= Running the {name-mission-secured} Booster Integration Tests on {VertX}
+[#running-the-secured-booster-integration-tests-on-{parameter-runtime}]
+= Running the {name-mission-secured} Booster Integration Tests on {runtime}
 
 .Prerequisites
 
 * The `oc` client authenticated.
+// * An empty OpenShift project in which to run the tests.
 
 .Procedure
 
@@ -15,13 +21,26 @@
 --
 oc create -f service.sso.yaml
 --
++
 . Wait until the {RHSSO} server is ready. Go to the Web console or view the output of `oc get pods` to check if the pod running the {RHSSO} server is ready.
 . Deploy the Secured HTTP API application:
 +
+ifdef::secured-vertx-mission[]
+
 [source,bash,options="nowrap",subs="attributes+"]
 --
 mvn fabric8:deploy -Popenshift
 --
+endif::secured-vertx-mission[]
+
+ifdef::secured-spring-boot-mission[]
++
+[source,bash,options="nowrap",subs="attributes+"]
+--
+$ mvn spring-boot:run -DSSO_AUTH_SERVER_URL=$(oc get route secure-sso -o jsonpath='{"https://"}{.spec.host}{"/auth\n"}')
+--
+endif::secured-spring-boot-mission[]
+
 . Wait until the Secured HTTP API is ready. Go to the Web console or view the output of `oc get pods` to check if the pod running the HTTP API endpoint is ready.
 . Execute the integration tests:
 +
@@ -29,4 +48,3 @@ mvn fabric8:deploy -Popenshift
 --
 mvn verify -Popenshift-it
 --
-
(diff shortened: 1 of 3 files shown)


docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
```diff
@@ -1,32 +1,57 @@
...
 * The `oc` client authenticated.
+// Should probably include the prerequisite to create a separate project for IT's in OpenShift
 
 .Procedure
 
+ifdef::secured-spring-boot-mission[]
+
+[IMPORTANT]
+--
+Keep the contents of the `service.sso.yaml` and the `.openshiftio/service.sso.yaml` templates synchronized to ensure that the integration tests work correctly.
```
Ladicek: Do we actually expect end users to modify these files? This is just a note for booster authors I believe.

inoxx03: @Ladicek You're right. I don't think users need to modify them in any way. If you see no reason for us to keep the not, I'm ok with removing it.

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
```diff
@@ -1,32 +1,59 @@
 
-[#running-the-secured-booster-integration-tests-on-vertx]
-= Running the {name-mission-secured} Booster Integration Tests on {VertX}
+[#running-the-secured-booster-integration-tests-on-{parameter-runtime}]
+= Running the {name-mission-secured} Booster Integration Tests on {runtime}
+//TODO: specify parameter in master file
```
tradej: Please remove :)

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
```diff
@@ -1,32 +1,59 @@
...
 
 .Prerequisites
 
+
 * The `oc` client authenticated.
+// Should probably include the prerequisite to create a separate project for IT's in OpenShift
```
tradej: If the process requires it, I certainly concur.

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
```diff
@@ -1,32 +1,59 @@
...
+////
+ifdef::secured-spring-boot-mission[]
+
+[IMPORTANT]
+--
+Keep the contents of the `service.sso.yaml` and the `.openshiftio/service.sso.yaml` templates synchronized to ensure that the integration tests work correctly.
+--
+
+endif::secured-spring-boot-mission[]
+////
```
tradej: Please remove.

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
inoxx03: Ah, thanks, will do:)

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
inoxx03: IIRC, it used to be the case that running integration tests inside an OpenShift project already containing a booster application resulted in the removal of this pre-existig application once the test were done. If that still apllies, using a separate project to run the IT's could  help  avoid accidental deletion of a running application. We could either include a step to create an extra project or include a note about it to alert the users to this possibility. @Ladicek WDYT?

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
```diff
@@ -1,10 +1,11 @@
 
-[#running-the-secured-booster-integration-tests-on-vertx]
-= Running the {name-mission-secured} Booster Integration Tests on {VertX}
+[#running-the-secured-booster-integration-tests-on-{parameter-runtime}]
```
tradej: Sorry, I forgot, can you please add the following comment to the file?

```
// This is a parameterized module. Parameters used: 
// 
//   parameter-$SOMETHING: $DESCRIPTION_OF_THE_PARAMETER
//   parameter-$SOMETHING_ELSE: ...
//
// Rationale: $YOUR_RATIONALE
```

I will merge it then.

docs/topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
Ladicek: So, if all the boosters migrated their tests to Arquillian Cube, there would be a very simple answer: verify that you have permissions to create new projects in OpenShift and add `-Dnamespace.use.current=false -DenableImageStreamDetection=true` when running the tests. With these parameters, Arquillian Cube would automatically create an ephemeral OpenShift project, run the tests in it and then destroy it.

Unfortunately, some boosters still use OpenShiftTestAssistant for their tests :-/

I think adding a note about the possibility to run the tests in an extra namespace is a good idea.

inoxx03:
@tradej added the comment explaining parameters, please let me know what you think.

tradej:
Cool, thanks. Merging...

tradej:
merged: 8036ac6a2

Issue #773, opened 2017-11-06
Clarify wording for ConfigMap mission


metacosm:
ConfigMap is both an API and Kubernetes objects. So we use the ConfigMap API but we create *a* ConfigMap, interact with ConfigMaps, etc.

rhoads-zach:
fixed in #774

PR #774, opened 2017-11-06
#773: clarify wording

metacosm:


docs/topics/configmap-mission-intro-paragraph.adoc
```diff
@@ -4,4 +4,4 @@ ifdef::docs-topic[xref:proficiency_foundational[*{proficiency-foundational}*].]
 ifndef::docs-topic[link:https://appdev.openshift.io/docs/wf-swarm-runtime.html#proficiency_levels[*{proficiency-foundational}*^].]
 
 
-The {name-mission-configmap} Mission provides a basic example of using ConfigMap to take advantage of external configuration sources.
+The {name-mission-configmap} Mission provides a basic example of using the ConfigMap API to take advantage of external configuration sources.
```
Ladicek: This is not true with WildFly Swarm and in general. ConfigMaps can be consumed in a bunch of ways: exposed to the container as a volume mount, exposed to the pod as env variables, or using the OpenShift API.

tradej:
ok to test

docs/topics/configmap-mission-intro-paragraph.adoc
metacosm: I guess that is confusing. I meant ConfigMap *API* more in a generic way. We can replace "the ConfigMap API" by ConfigMap*s* as opposed to just ConfigMap, which to me, doesn't mean much.

Ladicek: I'd probably reword completely. Instead of "using ConfigMap to take advantage of external configuration sources", I'd say something like "using a ConfigMap to externalize configuration".

rhoads-zach:
LGTM, merging.

rhoads-zach:
merged: c8111623a1e4c5942c774a74995343d4cc029a5c

Issue #772, opened 2017-11-03
remove note about tech preview from Vert.x runtime details


inoxx03:
[Vert.x Runtime Details](https://appdev.openshift.io/docs/vertx-runtime.html#_runtime_details)

rhoads-zach:
fixed in #804

PR #804, opened 2017-11-07
Issue #772: updated module with runtime-specific info

inoxx03:
Cleanup + remove notes about tech preview

### DIFF
--- a/docs/topics/vertx-runtime-info.adoc
+++ b/docs/topics/vertx-runtime-info.adoc
@@ -1,0 +1,38 @@
+= {Vertx} Runtime Information
+
+.Certified {Vertx} Version
+
+{Vertx} version `3.4.2-redhat-006` is certified to run on OpenShift and provided as part of the Red Hat subscription.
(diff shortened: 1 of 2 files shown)


docs/topics/vertx-runtime-info.adoc
```diff
@@ -0,0 +1,38 @@
+= {Vertx} Runtime Information
+
+.Certified {Vertx} Version
+
+{Vertx} version `3.4.2-3` is certified to run on OpenShift and provided as part of the Red Hat subscription.
```
Ladicek: Where does this version number come from? I think I commented about this somewhere already -- the Vert.x GA version number is `3.4.2.redhat-006`.

inoxx03: You're right. This version is not necessarily correct, as the changes introduced in this PR have not yet been reviewed by the Vert.x.
Thank you for catching this, I'll update momentarily:)

rhoads-zach:
merged: e42f6eba23e38fc060a4f96b64230f7adcc208a7

PR #817, opened 2017-11-07
incorporated additional feedback to gsg

rhoads-zach:
@cmoulliard @ALRubinger @jclingan mind taking a look?

### DIFF
--- a/docs/getting-started/master.adoc
+++ b/docs/getting-started/master.adoc
@@ -8,6 +8,12 @@ include::topics/templates/document-attributes.adoc[]
 = {getting-started-guide-name}
 
 [.lead]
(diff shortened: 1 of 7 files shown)


ALRubinger:
LGTM to merge.  I subsequent rounds I wonder if we can also pay some specific attention to the opening pieces to make them read more simply.  Everything here looks correct to me, if a bit verbose.  And I do like that we're adding it and appreciate that Zach went over this so cleanly. 

Concision is an art we can iterate on later. :)

docs/getting-started/master.adoc
```diff
@@ -8,6 +8,12 @@ include::topics/templates/document-attributes.adoc[]
 = {getting-started-guide-name}
 
 [.lead]
+With the emergence of Linux containerization and container platforms, the barrier to entry for using a polyglot, or multiple application runtime, architecture has been reduced. This allows you to embrace different ecosystems, languages, and runtimes at the same time for solving business problems. Moreover container platforms such as OpenShift manage, secure, and automate the deployments and monitoring of your applications. This enables you to decompose your business problems into smaller microservices while letting OpenShift help with managing some of the inherit complexity that comes with deploying, monitoring, and maintaining microservices.
```
Ladicek: I guess you meant "inherent complexity" instead of "inherit"?

PR #876, opened 2017-11-10
fixes #848

### LINKED ISSUE #848: incorrect identifiers in Swarm "4.6.7. Interacting with the Secured Booster"
This chapter uses `secured-wf-http` as an identifier of the route/service, but that's wrong. The correct one is `wfswarm-rest-http-secured`.

rhoads-zach:
@tradej @metacosm @Ladicek let me know what you think

### DIFF
--- a/docs/topics/proc_getting-the-secured-booster-api-endpoint.adoc
+++ b/docs/topics/proc_getting-the-secured-booster-api-endpoint.adoc
@@ -33,10 +33,10 @@ A sample output is shown in the following table:
 | <all>
 | passthrough
 
-| secured-{parameter-runtime}-http
-| secured-{parameter-runtime}-http-myproject.LOCAL_OPENSHIFT_HOSTNAME
+| PROJECT_ID
+| PROJECT_ID-myproject.LOCAL_OPENSHIFT_HOSTNAME
 |
-| secured-{parameter-runtime}-http
+| PROJECT_ID
 | <all>
 |
 
@@ -50,3 +50,4 @@ A sample output is shown in the following table:
 ====
 --
 
+IMPORTANT: `PROJECT_ID` is based on the name you entered when generating your booster using link:{link-launcher-oso}[{launcher-oso}] or the {launcher} tool. The default value for `PROJECT_ID` is `{parameter-runtime}-rest-http-secured`.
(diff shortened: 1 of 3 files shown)


docs/topics/proc_getting-the-secured-booster-api-endpoint.adoc
```diff
@@ -33,10 +33,10 @@ A sample output is shown in the following table:
 | <all>
 | passthrough
 
-| secured-{parameter-runtime}-http
-| secured-{parameter-runtime}-http-myproject.LOCAL_OPENSHIFT_HOSTNAME
+| {parameter-runtime}-rest-http-secured
```
metacosm: if `{parameter-runtime}` is `spring-boot` for the Spring Boot runtime, then yes. Hmm, on second thought, maybe not. This value is actually the artifact Id of the Maven project. While this value is correct for the Spring Boot runtime  booster *before* Launcher takes it and "clones" it with the user having the possibility to change the artifact id. So in fact, we don't know what that value may be until runtime…

rhoads-zach: in this case, the parameter is "sb".

what if we changed the whole thing to <NAME_OF_ARTIFACT> or something similar?

metacosm: `sb` is not correct, then. But in any case, this needs to be changed. The problem with using <NAME_OF_ARTIFACT> is that we don't currently have no explanation of the launch process to reference and explain that the artifact id is what was chosen when the booster was created in Launch (or the default value generated by Launch, which is *not* the value that we set in the boosters)… This needs to be well documented and it is not trivial to do so.

rhoads-zach:
@metacosm updated with generic value and explanation. Let me know what you think.

metacosm:
LGTM

rhoads-zach:
thanks @metacosm, merging.

docs/topics/proc_getting-the-secured-booster-api-endpoint.adoc
```diff
@@ -50,3 +50,4 @@ A sample output is shown in the following table:
 ====
 --
 
+IMPORTANT: `PROJECT_ID` is based on the name you entered when generating your booster using link:{link-launcher-oso}[{launcher-oso}] or the {launcher} tool. The default value for `PROJECT_ID` is `{parameter-runtime}-rest-http-secured`.
```
Ladicek: The concept of "default value" IMHO doesn't make a lot of sense to me here, and the value isn't actually correct :-) See https://github.com/openshiftio-vertx-boosters/vertx-secured-http-booster/blob/master/pom.xml#L29

PR #879, opened 2017-11-13
Fixed WFS additional resources links

### LINKED ISSUE #873: Additional resources are pointing to weird versions
Investigate and fix the versions of the WFS upstream docs we're pointing to.

* The HowTo and Reference Guides are pointing to the latest snapshot
* The User Guide points to 2017.6.0

tradej:
Resolves #873.

### DIFF
--- a/docs/topics/wf-swarm-dev-guide-additional-resources.adoc
+++ b/docs/topics/wf-swarm-dev-guide-additional-resources.adoc
@@ -1,8 +1,8 @@
 [[wf-swarm-additional-resources]]
 = Additional Resources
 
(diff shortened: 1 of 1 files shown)


Ladicek:
I don't quite understand why we're linking to community docs now that everything relevant from community docs is also present in product docs?

Oh wait, we don't have the HOWTOs, right? But I'm pretty sure we have the reference and most likely the user guide too.

tradej:
We don't have _all_ the howtos, for example creating a Datasource etc.

IMHO, the links to the upstream are useful for the user.

rhoads-zach:
merging.

rhoads-zach:
merged: efb675c61345cca4439497a433fb0a3d8e1da9f3

PR #886, opened 2017-11-14
Issue 792: update circuit breaker interaction

### LINKED ISSUE #792: Clarify circuit breaker fail response scenario
In section [2.5.6. Interacting with the Unmodified Spring Boot Tomcat Booster](https://doc-stage.usersys.redhat.com/documentation/en-us/red_hat_openshift_application_runtimes/1/html-single/spring_boot_tomcat_runtime_guide/#interacting_with_the_unmodified_spring_boot_tomcat_booster_4) when simulating a name service failure there is a following statement:

> If the remote endpoint is unavailable, it responds with an HTTP code 500 and you receive a predefined fallback response:

In might be worth pointing out that remote endpoint which returns status code 500 is `name` service. It might be confusing for the user because he is also calling a remote endpoint (`greeting`) which for him will return status code 200 and a body `{"content":"Hello, Fallback!"}`.

Also, the following statement in the same section is not accurate:

> The Circuit Breaker state indicator in the web interface changes from CLOSED to OPEN and the Circuit Breaker issues the fallback response when you invoke the /api/greeting endpoint.

Circuit is opened only when a failure threshold is reached, thus multiple requests to the failing service are needed before the circuit is opened and state indicator is updated. Same is true in the opposite scenario from OPEN to CLOSED.

inoxx03:
fixes #792 
does not affect  the Node guide, (node is covered by #887)

@gytis Thoughts?

### DIFF
--- a/docs/spring-boot-runtime/master.adoc
+++ b/docs/spring-boot-runtime/master.adoc
@@ -274,11 +274,13 @@ include::topics/assembly_building-and-deploying-the-booster-to-openshiftcontaine
 
 
 // CB basic interaction
-include::topics/circuit-breaker-mission-basic-interaction-spring-boot-tomcat.adoc[leveloffset=+3]
+:parameter-runtime-name: {SpringBoot}
+:context: {parameter-mission}-spring-boot
+include::topics/proc_interacting-with-the-circuit-breaker-booster.adoc[leveloffset=+3]
+
 //TODO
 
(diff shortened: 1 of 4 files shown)


gytis:
+1

rhoads-zach:
merged: 9847b709733ec071eedb3aa76b3742959b18ec82

PR #883, opened 2017-11-13
fixed dead xref

rhoads-zach:

PR #890, opened 2017-11-14
crud mision xref now points to crud IT section

inoxx03:
The xref in the crud mission info pointed to the configmap integration tests section.
Fixed by moving the relevant parameters before the include statement for the CRUD intro module in the master.adoc files for VX, SB & WF. 
To avoid definig the same values and parameters multiple times per mission, these parameters have been deduplicated and are defined at the top of the section, so that they can be properly consumed by all of the included parameterized modules.


inoxx03:
this is a follow-up to  pr #883

rhoads-zach:
merged: 507fe2768557d93169a628d99f60da905c8fe3d9

PR #894, opened 2017-11-14
updated federated ID callout

rhoads-zach:
@Ladicek @ALRubinger let me know what you think

### DIFF
--- a/docs/topics/proc-getting-started-create-booster.adoc
+++ b/docs/topics/proc-getting-started-create-booster.adoc
@@ -18,9 +18,13 @@ You can quickly create, build, and deploy a booster to {OpenShiftOnline} using l
 ** You may be prompted to log in to your Red Hat Developers account. If you are, click _Log in or register_ and complete the authentication steps.
 . You may be prompted to grant {launcher-oso} access to your GitHub and {OpenShiftOnline} accounts. This is a one-time action.
 +
-[WARNING]
+[IMPORTANT]
 ====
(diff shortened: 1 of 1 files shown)


Ladicek:
Copying my e-mail to here, too:

What I wanted to express is that the realm brings all these 3 accounts together. I'm not sure if "connect together" is proper English, but I'd myself write it like this:

> developers.redhat.com/launch uses the rh-developers-launch SSO realm to connect together your Red Hat Developers account, your GitHub account, and your OpenShift Online account. When you add these connections, developers.redhat.com/launch is granted access to those accounts and is allowed to use those accounts when creating and deploying boosters.
>
> When connecting these three accounts together in rh-developers-launch, they may only be used with each other within the rh-developers-launch SSO realm. For example, if your GitHub account is associated with a different Red Hat Developers account within the rh-developers-launch SSO realm, you must deauthorize it from the other Red Hat Developers account or add a different GitHub account.
>
> At this time, you cannot use the us-east-2 OpenShift Starter cluster with developers.redhat.com/launch. This cluster is used for OpenShift.io and is configured differently.

rhoads-zach:
Fixed @Ladicek, merging.

rhoads-zach:
merged: 371aea7b438a2514ad11c74a4e402de60b65dfb6

Ladicek:
Thanks @rhoads-zach, hopefully everyone else is fine with this text too :-)

PR #917, opened 2017-11-22
WIP - DO NOT MERGE: Fixes #912: update WFS version

### LINKED ISSUE #912: WildFly Swarm guide: missing product version number
The Swarm guide doesn't tell what's the version number to use when adding Maven dependencies. What's more, when showing the BOM usage, it even uses a community version number for demonstration!

FTR, the correct version number for our first GA is `7.0.0.redhat-8`.

inoxx03:
@Ladicek I've updated the WildFly Swarm version attribute to show the Productized version. WDYT?

### DIFF
--- a/docs/topics/templates/document-attributes.adoc
+++ b/docs/topics/templates/document-attributes.adoc
@@ -82,6 +82,11 @@
 
 :MinishiftVersion: 1.7.0
 :CDKVersion: 3.1.1
+
+// used in the BOM file example. 
+:WildFlySwarmProductVersion: 7.0.0.redhat-8
+
+// used in the links to additional upstream Swarm docs.
 :WildFlySwarmVersion: 2017.10.0
 
 :link-http-api-level-0-spring-boot-tomcat-booster: https://github.com/snowdrop/rest_springboot-tomcat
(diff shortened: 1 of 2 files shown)


Ladicek:
This is wrong IMHO. One, it doesn't affect the WildFly Swarm guide, and two, it's used in URLs to Swarm community docs.

inoxx03:
@Ladicek I believe the change does indeed affect the WildFly Swarm guide (Specifically the chapter that deals with using the BOM). The version displayed in the guide is sourced from the variable that has been changed.

I have completely missed the URLs, though, so thank you for pointing that out:) I'll fix those and update the PR.

Ladicek:
I think this breaks the rules of how WildFly Swarm docs should be updated, but is OK for now. @tradej will need to fix this properly when he returns :-)

rhoads-zach:
thanks @Ladicek.
merging

rhoads-zach:
merged: 999fc44ec7a6e3dc2b63d8b84b284d19f9c7e94f

inoxx03:
@lthon Thank you:) FTR, I'd personally be more comfortable letting Sheldon handle this, too... But as an emergency fix, I hope this'll do:)

tradej:
Hi, just to confirm, I believe this fix is perfectly fine and can stay as a permanent one. We both want the product version (now stored in `WildFlySwarmProductVersion`) and the community version (now stored in `WildFlySwarmVersion`, but it can be renamed to, say, `WildFlySwarmCommunityVersion`).

As for the rules of updating, since no files in `docs/topics/wildfly-swarm` were changed, this PR is 100% by the book.

As for setting the WFS product version in `document-attributes.adoc`, I'm afraid we'll have to do it because we can not reproduce a WFS product build completely (setting the version happens there), so this is acceptable.

Ladicek:
I knew there is a directory with `wildfly-swarm` in the name that must not be updated manually, and probably some brain hiccup caused me to believe that this PR did that. Happy to have been wrong :-)

tradej:
I appreciate you remembered that; it's better than if it was forgotten. :)

PR #968, opened 2017-12-18
Added the -r parametre to jq usage in SSO mission

### LINKED ISSUE #911: secured boosters -- use `jq -r` to get rid of quotes
The secured boosters docs instruct users to

> `curl ... | jq '.access_token'`

and then warn

> When using this result, make sure to remove the surrounding double quotes (").

Instead, we should tell users to `curl ... | jq -r '.access_token'`, which avoids the quotes altogether. Quoting from `man jq`:

>        ·   --raw-output / -r:
>
>           With this option, if the filter´s result is a string then it will be written directly to standard output rather than being formatted as a JSON string with
>           quotes. This can be useful for making jq filters talk to non-JSON-based systems.

tradej:
Resolves #911.


Ladicek:
+1000

rhoads-zach:
merged: 773d6423cb5f206e1fc97e969ad198b5c1d73165

PR #970, opened 2017-12-18
Fixes #775 updated configmap intro

### LINKED ISSUE #775: It would make sense to explain ConfigMap before describing the mission
i.e. move About ConfigMap section from configmap-mission-intro to configmap-mission-intro-paragraph?

lavernw1:
@metacosm @rhoads-zach @tradej moved configmap definition to intro paragraph for externalized configuration mission.

### DIFF
--- a/docs/topics/configmap-mission-intro.adoc
+++ b/docs/topics/configmap-mission-intro.adoc
@@ -11,6 +11,6 @@ It is important for the application configuration to be externalized and separat
 Microservices architectures and polyglot environments add a layer of complexity to managing an application's configuration. Applications are comprised of independent, distributed services, each potentially with its own configuration.
 This creates a maintenance challenge to keep the configuration synchronized and accessible from all services.
 
-ConfigMaps enable the application configuration to be externalized and used in individual Linux containers and pods in a language-independent way on OpenShift. You can create a ConfigMap object in a variety of ways, including using a YAML file, and inject it into the Linux container.
-ConfigMaps also allow sets of configuration data to be easily grouped and scaled, which enables you to configure an arbitrarily large number of environments beyond the basic _Development_, _Stage_, and _Production_.
-You can find more information about ConfigMap in the link:https://docs.openshift.org/latest/dev_guide/configmaps.html[OpenShift documentation].
+ConfigMaps enable the application configuration to be externalized and used in individual Linux containers and pods on OpenShift. You can create a ConfigMap object in a variety of ways, including using a YAML file, and inject it into the Linux container.
+ConfigMaps also allow sets of configuration data to be easily grouped and scaled. This lets you configure an arbitrarily large number of environments beyond the basic _Development_, _Stage_, and _Production_.
+You can find more information about ConfigMaps in the link:https://docs.openshift.org/latest/dev_guide/configmaps.html[OpenShift documentation].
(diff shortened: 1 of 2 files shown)


tradej:
If I am not mistaken, that intro file is used by the launcher. @gastaldi, can you please verify this doesn't break anything on your end?

gastaldi:
The files used by launcher are listed in https://github.com/fabric8-launcher/launcher-frontend/blob/master/src/assets/adoc.index.

The only affected file would be configmap-mission-intro-paragraph.adoc afaik

tradej:
> This makes the UI a bit weird IMHO, so I am not sure we want that change

That's what I was afraid of.

lavernw1:
@gastaldi @tradej  Thanks for your comments.  
I propose restoring the configmap-mission-intro-paragraph.adoc, and to only include the sentence with the definition of a configmap from the About Configmap section.  The change would be as follows: WDYT?

Mission proficiency level: Foundational.

The Externalized Configuration Mission provides a basic example of using a ConfigMap to externalize configuration.  A ConfigMap is an object used by OpenShift to inject configuration data as simple key and value pairs into one or more Linux containers while keeping the containers agnostic of OpenShift.

gastaldi:
@lavernw1 that looks like a much better idea. I'm ok with that if everybody agrees

lavernw1:
@gastaldi thanks

tradej:
I'm not a big fan of using the word _agnostic_ since it's a farily technical term in a chunk of very basic text that should be comprehensible for everybody. Can we replace it with something else, e. g. "independent" or the like?

Otherwise, this solution is acceptable for me.

lavernw1:
@tradej - Thanks I'll change the wording to  "independent of OpenShift "

lavernw1:
@gastaldi @tradej @rhoads-zach updated with the suggested changes.

lavernw1:
@gastaldi is there a problem with the C/I. Build failing.

gastaldi:
Retest this please

docs/topics/configmap-mission-intro.adoc
```diff
@@ -8,11 +9,10 @@ This mission shows you how to:
 .About Externalized Configuration
 It is important for the application configuration to be externalized and separated from its code. This allows the application configuration to change as it moves through different environments while leaving the code unchanged. This also keeps sensitive or internal information out of your code base and version control. Many languages and application servers provide environment variables to support externalizing an application's configuration.
 Microservices architectures and polyglot environments add a layer of complexity to managing an application's configuration. Applications are comprised of independent, distributed services, each potentially with its own configuration.
-This creates a maintenance challenge to keep the configuration synchronized and accessible from all services. 
+This creates a maintenance challenge to keep the configuration synchronized and accessible from all services.
 
 .About ConfigMap
```
metacosm: This should be renamed since it's not explaining what a ConfigMap is anymore. Maybe remove the sub-sections entirely and merge all the text together…  saying something to the effect of: ConfigMaps are OpenShift's solution to  externalizing configuration… Here's more info (link)… This is important because… (text from Why ConfigMap is important subsection).

metacosm:
LGTM

rhoads-zach:
LGTM, merging.

rhoads-zach:
merged: 9bed57ed50c51609a3c0e1b018d09aaa9b7c7337

PR #983, opened 2018-01-11
Removed note prohibiting sync with WFS prod repo + synced WFS repo

tradej:
Notes:

* It is now OK to sync the repository according to the latest Contribution Guide instructions.
  * The changes in the guide are not published now, so a re-publishing is in order.
* The WFS sources are now properly synchronized with the WFS prod repository.
  * Some fractions had wrong package name elements (e. g. `c-d-i` instead of `cdi`). That was a bug in the WFS repository that was fixed, and is now synced here.
  * There is a significant number of removed files. That is to be expected because originally, both a community and a product build was done, leaving many community fraction files in the repository. Now, we only do the product build, so the community fraction files are not created in the first place.

Resolves [OBST-550](https://issues.jboss.org/browse/OBST-550).


rhoads-zach:
merged: 666b1e9beb3eb795a49cba283b3d8059f0fe73a4

PR #984, opened 2018-01-12
fixes #926: add note about database login credentials

### LINKED ISSUE #926: Relation db login credentials
In springBoot doc 2.3.1 - relational database -> deployment contains command:
oc new-app -e POSTGRESQL_USER=luke -ePOSTGRESQL_PASSWORD=secret -ePOSTGRESQL_DATABASE=my_data openshift/postgresql-92-centos7 --name=my-database

are all the credentials and db name somehow significant? Does a user need to use them somewhere? The docs should answer this.

inoxx03:
@Ladicek WDYT? is this enough detail, or do you think we need to provide more context?

### DIFF
--- a/docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
+++ b/docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
@@ -31,6 +31,9 @@ $ oc new-project {project-name}
 . Navigate to the root directory of your booster.
 
 . Deploy the PostgreSQL database to OpenShift.
+Ensure that you use the following values for user name, password, and database name when creating your database application.
+The booster application is pre-configured to use these values.
+Using different values prevents your booster application from integrating with the database.
 +
 [source,bash,options="nowrap",subs="attributes+"]
 ----
--- a/docs/topics/readme/crud-README.adoc
+++ b/docs/topics/readme/crud-README.adoc
@@ -16,12 +16,16 @@ $ oc login -u developer -p developer
 
 $ oc new-project MY_PROJECT_NAME
 
+# Ensure that you use the following values for the user name, password and database name when creating your database application.
+# The pre-configured values are used in the `credentials-secret.yml` and `deployment.yml` files in the `src/main/fabric8` directory of your booster application project.
+
 $ oc new-app -e POSTGRESQL_USER=luke -ePOSTGRESQL_PASSWORD=secret -ePOSTGRESQL_DATABASE=my_data openshift/postgresql-92-centos7 --name=my-database
 
 # Wait for `my-database` application to be running.
 
 $ ${OSORunCMD}
 ----
+
 
 == Interacting with the Booster on a Single-node OpenShift Cluster
 


docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
```diff
@@ -31,6 +31,8 @@ $ oc new-project {project-name}
 . Navigate to the root directory of your booster.
 
 . Deploy the PostgreSQL database to OpenShift.
+Ensure that you use the following login credentials when creating the database application.
```
Ladicek: Personally, I've never been a fan of the word "credentials", and it's not just credentials here, it's also the database name. So I'd replace "login credentials" with "username, password and database name".

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
```diff
@@ -31,6 +31,8 @@ $ oc new-project {project-name}
...
+Using different credentials might prevent your booster application from integrating with the database.
```
Ladicek: "Using different values will prevent your booster application from connecting to the database."

Ladicek: Maybe also add one more line saying something like "This is because the booster application is preconfigured to use these values." It would be possible to also list the files where each booster uses these values (I provided the list in #926).

docs/topics/readme/crud-README.adoc
```diff
@@ -16,13 +16,16 @@ $ oc login -u developer -p developer
 
 $ oc new-project MY_PROJECT_NAME
 
+# Ensure that you use the login credential values specified below when creating the database application.
```
Ladicek: Same as above, "login credential values" -> "username, password and database name".

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
inoxx03: @Ladicek I agree that 'credentials' might not be the most accurate term here. I'll make the change here.

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
inoxx03: Thank you:) IMHO that's definitely useful. I'll add that information in.

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
```diff
@@ -31,6 +31,9 @@ $ oc new-project {project-name}
 . Navigate to the root directory of your booster.
 
 . Deploy the PostgreSQL database to OpenShift.
+Ensure that you use the following values for user name, password, and database name when creating your database application.
+The booster application is pre-configured to use this values.
```
Ladicek: "these values"

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
```diff
@@ -31,6 +31,9 @@ $ oc new-project {project-name}
...
+Using different values might prevent your booster application from integrating with the database.
```
Ladicek: Might is IMHO wrong. It's not a possibility, it's certain.

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
inoxx03: Thanks, that one should be _these_:)

docs/topics/proc_deploying-the-crud-booster-using-the-oc-cli-client.adoc
inoxx03: Let me correct that. 
I'll also add information about the yAML files that contain the pre-configured username, password and db name

PR #995, opened 2018-01-24
Added initial API info. RHOARDOC-1145

rhoads-zach:
@lance @lholmquist thoughts?

### DIFF
--- a/docs/nodejs-runtime/master.adoc
+++ b/docs/nodejs-runtime/master.adoc
@@ -201,7 +201,12 @@ include::topics/secured-mission-resources.adoc[leveloffset=+3]
 include::topics/proc_debug-logging-nodejs.adoc[leveloffset=+2]
 
 
-//== Creating an Application from Scratch
+== Developing an Application for the {NodeJS} Runtime
+
+include::topics/con_nodejs-runtime-api.adoc[leveloffset=+2]
+
+include::topics/con_creating-nodejs-app.adoc[leveloffset=+2]
+
 //todo
 //Building blocks
 //Minimum Viable App
--- a/docs/topics/con_creating-nodejs-app.adoc
+++ b/docs/topics/con_creating-nodejs-app.adoc
@@ -1,0 +1,142 @@
+= Creating a {NodeJS} Application
+
+In addition to xref:mission-http-api-nodejs[using a booster], you can also create new {NodeJS} applications from scratch and deploy them to OpenShift.
+
+== Creating an Application
+
+.Prerequisites
+* `npm` installed.
+
+.Procedure
+
+. Create the application folder.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mkdir myApp
+----
+
+. Initialize your application with npm.
++
+The rest of this example assumes the entry point is `app.js`, which you are prompted to set when running `npm init`.
++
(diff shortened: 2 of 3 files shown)


docs/topics/proc_using-nodemon-for-dev.adoc
```diff
@@ -0,0 +1,63 @@
+[#using-nodemon-for-development]
+= Using nodemon for Development
+
+The {NodeJS} runtime includes link:https://nodemon.io/[nodemon^], which watches a directory and restarts your application when it sees a change. This is useful during
```
Ladicek: during what? :-)

docs/topics/proc_using-nodemon-for-dev.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+[discrete]
+== Prerequisites
+* An application you want to use with nodemon. For example, xref:mission-http-api-nodejs[a booster].
+* `npm` installed.
+
+[discrete]
+== Procedure
+
+. Use `npm` to install nodemon.
```
Ladicek: Above, you said that the Node.js runtime _includes_ nodemon. Does it really, if you need to install it?

Ladicek:
So, I don't know anything about `nodemon`, but it seemed useful. Why is it gone now? I'd be totally fine with the previous content, if it didn't tell that nodemon is included. (Well maybe it created an impression that it's somehow "supported" as part of RHOAR, but that's just about wordsmithing, right? :-) )

rhoads-zach:
@Ladicek i might re-introduce in the debugging section

rhoads-zach:
@lance @lholmquist thoughts on the the new section?

rhoads-zach:
@lholmquist @lance @Ladicek 

Re-added nodemon as an appendix

danbev:
> We also need to confirm with @danbev if the S2I builder image for production includes this.

`nodemon` is currently not included in the production S2I image. My take on this from the last discussion we had is that we'd document the steps to set it up but not provide anything except the node runtime itself in the docker image.

rhoads-zach:
@lance @danbev thank you for the feedback. I'll remove the nodemon content again and revisit at a later date.

FYI @Ladicek ^^

Ladicek:
I think the content we have right now is perfectly fine for local usage, and that's still valuable. Sure, you can have `nodemon` running in the OpenShift pod and use it by `oc rsync`-ing stuff (could you even use `nodemon` to trigger the `os rsync`?).

Anyway, I'm not insisting on keeping it right now; we can always revisit later.

rhoads-zach:
@lance removed nodemon for now and will revisit later. can you give a +1 if you're good with the current state?

lance:
@rhoads-zach I can, although we are still talking about this on the team and ways in which we can support it (not as official support, but just so the build doesn't break). Do you want to wait until we sort that out?

rhoads-zach:
@lance i think we should break the nodemon stuff out into a different issue so we can wrap this PR up. Do you have any objections to that? Re-adding the content is not hard.

rhoads-zach:
thank you @lance 

@inoxx03 @tradej feel free to merge.

inoxx03:
@rhoads-zach LGTM, merging...

inoxx03:
merged: c776dbdf0c079a4512cc1f55addd1979663674cc

PR #1002, opened 2018-01-31
added live debugging content, refactored debug logging. RHOARDOC-1142

rhoads-zach:
@lholmquist @lance mind taking a quick look?


docs/topics/proc_debug-start-native-debug-nodejs.adoc
```diff
@@ -0,0 +1,30 @@
...
+----
+$ node inspect app.js
+< Debugger listening on ws://127.0.0.1:9229/12345678-aaaa-bbbb-cccc-0123456789ab
+< For help see https://nodejs.org/en/docs/inspector
+< Debugger attached.
+...
+debug>
+----
++
+If you have an link:https://expressjs.com[express] application, such as the xref:mission-http-api-nodejs[{name-mission-http-api} booster], starting the debugger is slightly different:
```
lholmquist: I wonder if we should mention if you create the express application using the express-generator, https://expressjs.com/en/starter/generator.html

that is what creates the `./bin/www`   but you are referencing the http booster,  i'll leave it up to you :)

docs/topics/proc_add-debug-logging-nodejs.adoc
```diff
@@ -0,0 +1,36 @@
...
+  debug('name: '+name);
+  response.send({content: `Hello, ${name || 'World'}`});
+});
+----
+
+. Install the link:https://www.npmjs.com/package/debug[debug] module.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ npm install debug
```
Ladicek: Would it be better here to show that the `debug` package is present in `package.json`? I don't think you will `npm install` anything in an OpenShift pod, for example.

docs/topics/proc_debug-start-native-debug-nodejs.adoc
Ladicek: Yea I think we should rather describe that what you need to do is, in general, `node inspect path/to/app/entrypoint`, rather than selectively choosing one particular web framework and saying that it's slightly different, while the difference is just the location of the entrypoint.

docs/topics/proc_debug-start-v8-debug-nodejs.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+.Procedure
+. Start your application with the link:https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js[V8 inspector integration enabled].
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ node --inspect app.js
+----
++
+If you have an link:https://expressjs.com[express] application, such as the xref:mission-http-api-nodejs[{name-mission-http-api} booster], starting the debugger is slightly different:
```
Ladicek: Same as above -- the only difference is the location of the entrypoint.

docs/topics/proc_show-debug-logging-local-nodejs.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+.Procedure
+. Set the `DEBUG` environment variable to enable debug logging in your application.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ DEBUG=mybooster npm start
+----
++
+The `DEBUG` environment variable uses link:https://www.npmjs.com/package/debug#wildcards[wildcards] to filter debugging messages.
```
Ladicek: I think it's the `debug` package who _uses_ wildcards. The `DEBUG` environment variable _can include_ wildcards :-)

docs/topics/proc_show-debug-logging-openshift-nodejs.adoc
```diff
@@ -0,0 +1,88 @@
...
+$ curl $APPLICATION_ROUTE/api/greeting?name=Sarah
+----
+
+. Return to your pod logs and notice there are no debug logging messages in the logs. 
+
+. Set the `DEBUG` environment variable in your application pod to enable debug logging.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ oc set env pod/POD_NAME DEBUG=mybooster
```
Ladicek: Does this work? It didn't for me, which is why the Java-based runtimes debugging docs (which also need to set an environment variable) change the entire deployment config. Changing a single pod would be much nicer! I need to confirm.

docs/topics/proc_add-debug-logging-nodejs.adoc
lance: I agree that the `package.json` excerpt is good here. It's also common to show the `install` command. However, it should be shown as `npm install --save debug` which will add the dependency to `package.json` for you.

docs/topics/proc_add-debug-logging-nodejs.adoc
```diff
@@ -0,0 +1,36 @@
...
+----
++
+Depending on your application, this module may already be included. For example, link:https://expressjs.com[express] applications link:https://expressjs.com/en/guide/debugging.html[already include this module], therefore you do not need to install it.
```
lance: I'm not sure I would include this text. It puts the burden on the user to figure out which packages include it and which don't. And even if some do, it will not hurt to install it.

docs/topics/proc_show-debug-logging-local-nodejs.adoc
lance: Slightly more refined, "The `DEBUG` environment variable _specifies_ which debug statements to print, and may include wildcards."

docs/topics/proc_show-debug-logging-openshift-nodejs.adoc
```diff
@@ -0,0 +1,88 @@
...
+----
+
+. Deploy your application to OpenShift.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ npm run openshift
+----
++
+This runs the `openshift` npm script, which wraps direct calls to xref:about-nodeshift[nodeshift].
```
lance: This is specified in the `package.json` file. As above, it might be good to include an excerpt here.

docs/topics/proc_show-debug-logging-openshift-nodejs.adoc
lance: I believe it should work, yes, since a config change like setting an env variable will cause a new pod to spin up and the only thing that determines whether debugging output is shown is that env var.

rhoads-zach:
thank you for all the feedback @lance @Ladicek @lholmquist 

Updated.

Let me know what you think.

docs/topics/proc_add-debug-logging-nodejs.adoc
```diff
@@ -0,0 +1,47 @@
...
++
+[source,json,options="nowrap",subs="attributes+"]
+----
+...
+"dependencies": {
+    "debug": "^3.1.0"
+  }
+----
++
+Depending on your application, this module may already be included. For example, when using the link:https://expressjs.com/en/starter/generator.html[express generator] to create your application, the `debug` module is already added to `package.json`. Some of the boosters, such as the xref:mission-http-api-nodejs[{name-mission-http-api} booster], use already have the `debug` module in the `package.json`.
```
lance: Typo: 'use already have'

rhoads-zach:
thanks @lance.

Fixed.

rhoads-zach:
@Ladicek Take a quick look and let me know if you want any more changes. 

If not, i'll merge first thing friday.

inoxx03:
merged: 754adb64b985976faf2611d8b9fbaa1d0f9a7e30

PR #1021, opened 2018-02-14
fixed build issues with WFS guide

rhoads-zach:


Ladicek:
I guess some of these fixes should be brought upstream?

rhoads-zach:
@Ladicek yes, was going to tag @tradej to back port these fixes

PR #1028, opened 2018-02-22
WIP: fixes RHOARDOC-1167: removed reference to kubernetes API in table

rhoads-zach:
@metacosm let me know what you think.


docs/topics/sb-features-and-frameworks-summary.adoc
```diff
@@ -46,10 +46,10 @@ a|
 Discovers Service/Endpoint deployed on OpenShift and exposed behind a service or route using the service name matching a DNS entry.
 
 a|
-Yes - using Kubernetes API
+Yes
```
Ladicek: I think this reference to the Kubernetes API might actually be correct (not sure), it's the first one (in the Circuit Breaker section) that I can't possibly see being right.

docs/topics/sb-features-and-frameworks-summary.adoc
```diff
@@ -20,10 +20,10 @@ a|
 Switches between services and continues to process incoming requests without interruption in case of service failure.
 
 a|
-Yes - using Kubernetes API
+Yes
 
 a|
-Spring Cloud Kubernetes - Hystrix
+link:https://github.com/spring-cloud-incubator/spring-cloud-kubernetes[Spring Cloud Kubernetes] - Hystrix
```
metacosm: Actually, this isn't using Spring Cloud Kubernetes, but rather Spring Cloud Netflix (https://github.com/spring-cloud/spring-cloud-netflix) Hystrix support (via the spring-cloud-starter-netflix-hystrix starter). Note that this starter has been renamed in more recent Spring Cloud version and this is the new name. Note also that we probably don't need to mention the actual starter for simplicity's sake.

docs/topics/sb-features-and-frameworks-summary.adoc
metacosm: Yes, discovery is also done using Kubernetes API so we can keep that one.

docs/topics/sb-features-and-frameworks-summary.adoc
Ladicek: Ha, very good point! I totally missed that.

rhoads-zach:
thanks for the feedback @metacosm and @Ladicek 

Addressed your feedback, let me know what you think.

docs/topics/sb-features-and-frameworks-summary.adoc
```diff
@@ -75,7 +75,7 @@ a|
 No
 
 a|
-Spring Cloud Kubernetes - Ribbon
+link:https://github.com/spring-cloud-incubator/spring-cloud-kubernetes[Spring Cloud Kubernetes] - Ribbon
```
Ladicek: Ribbon is actually also Netflix, right @metacosm?

Ladicek: OK, wrong, there's a Ribbon module in Spring Cloud Kubernetes. Not that I would understand it :-)

metacosm: I guess both would be correct, here… We're not using either in the boosters at the moment, though.

PR #1036, opened 2018-02-27
Debug logging fixes (RHOARDOC-1171)

tradej:
Things done:

* Renamed the debug logging–related files so that they follow the guidelines (even the localhost-related ones to keep consistency).
* Removed the separate WFS OpenShift debug logging chapter and reused the common one.
* Modified the common debug logging file so that it accommodates all three Maven-based runtimes.
  * The `:$RUNTIME-logging:` attribute was removed as it was not necessary (conditionalising by the existing `:wf-swarm:`, `:vert-x:`, etc. attributes is enough).
  * I took the liberty to do some stylistic and consistency edits, mostly taken from the original WFS file.

Rendered runtime guides:
* [Node.js](http://download.eng.brq.redhat.com/scratch/tradej/rhoar/RHOARDOC-1171/html/docs/nodejs-runtime.html#showing-debug-logging-on-openshift)
* [Spring Boot](http://download.eng.brq.redhat.com/scratch/tradej/rhoar/RHOARDOC-1171/html/docs/spring-boot-runtime.html#showing-debug-logging-on-openshift)
* [Vert.x](http://download.eng.brq.redhat.com/scratch/tradej/rhoar/RHOARDOC-1171/html/docs/vertx-runtime.html#showing-debug-logging-on-openshift)
* [WildFly Swarm](http://download.eng.brq.redhat.com/scratch/tradej/rhoar/RHOARDOC-1171/html/docs/wf-swarm-runtime.html#showing-debug-logging-on-openshift)

Resolves RHOARDOC-1171.


docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
+
+[#showing-debug-logging-on-openshift]
+= Showing Debug Logging on OpenShift
+
+Start your application and interact with it to see the debugging statements in OpenShift.
+
+.Prerequisites
+
+* A Maven-based application
+ifndef::wf-swarm[with debug logging enabled]
```
Ladicek: This conditional sounds wrong.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
...
+$ mvn clean fabric8:deploy -Popenshift
+----
+endif::[]
+
+ifdef::wf-swarm[]
+. Set the `swarm.logging` property to `DEBUG` when deploying your application to OpenShift:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean fabric8:deploy -Popenshift -Dswarm.logging=DEBUG
```
Ladicek: Did you verify? I don't this will do what you think it will do.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
...
+ifdef::spring-boot[]
+. To disable debug logging, remove `logging.level.fully.qualified.name.of.TheClass=DEBUG` from `src/main/resources/application.properties` and redeploy your application.
+endif::[]
+
+ifdef::vert-x[]
+. To disable debug logging, update your logging configuration file, for example `src/main/resources/vertx-default-jul-logging.properties`, remove the logging configuration for your class and redeploy your application.
+endif::[]
+
+ifdef::wf-swarm[]
+. To disable debug logging, redeploy your application without the `swarm.logging` property set:
```
Ladicek: Same as above. I'd recommend referring to `project-defaults.yml` here instead, similarly to Spring Boot and Vert.x above.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
...
+endif::[]
+ifdef::wf-swarm[]
+[source,options="nowrap",subs="attributes+"]
+----
+...
+Feb 11, 2017 10:23:42 AM io.openshift.{app-name}
+INFO: Greeting: Hello, Sarah
+...
+----
+endif::[]
```
Ladicek: Missing the Vert.x and Node.js parts.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: It's correct AFAICT; in WFS, I operate with the assumption that debug logging is not enabled, and use the `-Dswarm.logging=DEBUG` ad hoc setting, which IMO makes more sense for debugging as it is usually a temporary thing.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+ifdef::wf-swarm[]
+. To disable debug logging, redeploy your application without the `swarm.logging` property set:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean fabric8:deploy -Popenshift
+----
+endif::[]
```
Ladicek: Missing the Node.js part.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
```diff
@@ -0,0 +1,116 @@
...
+.Procedure
+
+ifndef::wf-swarm[]
+. Deploy your application to OpenShift:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean fabric8:deploy -Popenshift
+----
+endif::[]
```
Ladicek: This part is wrong for Node.js.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: I removed these intentionally as there are no nested ifdefs in the source file.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: Node.js does not use this file.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: Node.js does not use this file.

Ladicek:
One more general comment: "showing debug logging" sounds strange, there's too many gerunds. I'd call it "showing debug logs". In general, I'd say that "logging" is the process, and the results of that process are called "logs". That's probably outside of this PR scope, not sure how many places would be affected by such change.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: I'll check; why do you think it won't work, though? Won't the `swarm.logging` property be used in the deployment?

docs/topics/proc_showing-debug-logging-on-openshift.adoc
Ladicek: OK, so, well, I don't see the corresponding `endif` :-) Also I'd guess that WildFly Swarm also requires `oc` installed and authenticated.

tradej:
@Ladicek, I agree about the title, I'd change the titles of relevant chapters to "Accessing debug logs on [localhost|OpenShift]." WDYT?

cc @rhoads-zach

docs/topics/proc_showing-debug-logging-on-openshift.adoc
Ladicek: I'm 99.9% sure the system property won't propagate from your local machine to the pod inside the OpenShift cluster.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
Ladicek: OK, so if Node.js doesn't use this file then good, but Vert.x apparently does use this file, so should have an example here.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: If you include the conditional text in the square brackets, it's a one-liner without the `endif`.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: I see, I'll use to the YAML configuration then. Thanks.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
tradej: It didn't have one to begin with, so I'll create it.

Ladicek: Ah OK :-)

Ladicek:
"Accessing debug logs on [localhost|OpenShift]." sounds good to me.

docs/topics/proc_showing-debug-logging-on-openshift.adoc
Ladicek: Aaaaa :-) Sorry for the noise.

rhoads-zach:
@tradej LGTM. Do you want me to merge?

tradej:
@rhoads-zach Not yet, please. I still need to factor in the changes requested by Lada. Should be done today all right.

PR #1038, opened 2018-03-01
En dash usage (RHOARDOC-1177)

tradej:
**NOTE** Waiting for upstream WFS to accept the changes and sync them to the prod repo.

Changes:

* All explicit usage of `&#x2013;` is replaced with `{ndash}`, which is set to that value.
* Incorporates changes committed to the WFS upstream.
* Above practice is documented in the Contribution Guide.

Resolves RHOARDOC-1177.

### DIFF
--- a/docs/nodejs-runtime/master.adoc
+++ b/docs/nodejs-runtime/master.adoc
@@ -203,7 +203,7 @@ include::topics/secured-mission-resources.adoc[leveloffset=+3]
 
 
 == Debugging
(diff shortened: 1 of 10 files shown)


tradej:
Good to go.

I have set WFS to be ignored when validating because an unrelated commit to the upstream broke it (There's emphasis in a literal block, which is inadmissible in DocBook. However, it renders fine in AsciiDoc. I think we may need to reconsider the strictness of the validation.)

PR #1041, opened 2018-03-06
fix: node.js docs should show the usage of role-based auth

lholmquist:
goes along with this booster issue,  https://github.com/bucharest-gold/nodejs-rest-http-secured/issues/53


rhoads-zach:
LGTM, merging.

PR #1051, opened 2018-03-15
Updated node.js docs with instructions for using the 'files' section …

rhoads-zach:
…in package.json. Fixes RHOARDOC-1189

### DIFF
--- a/docs/topics/proc_creating-nodejs-app.adoc
+++ b/docs/topics/proc_creating-nodejs-app.adoc
@@ -102,6 +102,37 @@ $ npm install nodeshift --save
 +
 The `openshift` script uses `nodeshift` to deploy the application to OpenShift.
 
+. _Optional_: Add a `files` section in `package.json`.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+{
+  "name": "myApp",
+  "version": "1.0.0",
+  "description": "",
+  "main": "app.js",
+  "scripts": {
+    ...
+  },
+  "files": {
+    "package.json",
+    "app.js"
+  }
+  ...
+}
+----
++
+The `files` section tells `nodeshift` what files and directories to include when deploying to OpenShift.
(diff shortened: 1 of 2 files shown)


rhoads-zach:
@lholmquist let me know what you think

lholmquist:
👍

docs/topics/proc_creating-nodejs-app.adoc
```diff
@@ -102,6 +102,34 @@ $ npm install nodeshift --save
...
+  },
+  "files": {
+    "package.json",
+    "app.js"
+  }
+  ...
+}
+----
++
+The `files` section tells `nodeshift` what files to include when deploying to OpenShift. If the `files` section is not specified, `nodeshift` will send the entire current directory, excluding:
```
Ladicek: Multiple sentences on single line :-)

Also, more importantly -- can the entries in the `files` section contain wildcards? I'd expect to find out in this description. @lholmquist?

lholmquist: like how wild are you thinking.

if you want to do a directory, for example, `lib`, you can just specificy `lib` and that will get everything in that directory including sub directories.

this is the module we use to do that archiving: https://github.com/npm/node-tar
and we pass in an array of files to the create method.

if there is some crazy pattern matching stuff that someone needs, then i'd imaging they've written their app wrong

Ladicek: I think that's perfectly enough, but from the present text, I didn't get the impression that I can include directory names as well. I actually got the opposite impression -- that I need to list all the files.

lholmquist: that is true.  lets tell @rhoads-zach to add it  :)

rhoads-zach:
@lholmquist @Ladicek updated per feedback.  Let me know what you think.

lholmquist:
👍

rhoads-zach:
thank you @lholmquist @Ladicek 

merging.

PR #1054, opened 2018-03-20
Fixes RHOARDOC-1163

inoxx03:
procedure for building a Spring boot app from scratch

### DIFF
--- a/docs/topics/proc_creating-an-application.adoc
+++ b/docs/topics/proc_creating-an-application.adoc
@@ -1,0 +1,153 @@
+[#creating-an-application]
+= Creating an application
+
+Create a simple `Greeting` application to run on OpenShift using {SpringBoot}.
+The following procedure shows you how to:
+
+* Write some simple application code that makes use of functionalities provided by {SpringBoot}.
+* Declare dependencies and configure the application build using a `pom.xml` file.
+* Start  your application on localhost and verify that it works.
+
+
+.Prerequisites
+
+* Maven installed.
+
+.Procedure
+
+. Create the application directory and navigate to it.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mkdir myApp
+$ cd myApp
+----
+
+. Create a `pom.xml` file.
++
+[source,xml,options="nowrap",subs="attributes+"]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example</groupId>
+  <artifactId>my-app</artifactId>
+  <version>1.0.0-SNAPSHOT</version>
+
+  <name>MyApp</name>
+  <description>My Application</description>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>me.snowdrop</groupId>
+        <artifactId>spring-boot-bom</artifactId>
+        <version>1.5.10.SP4-SNAPSHOT</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-tomcat</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-actuator</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.cxf</groupId>
+      <artifactId>cxf-spring-boot-starter-jaxrs</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.jaxrs</groupId>
+      <artifactId>jackson-jaxrs-json-provider</artifactId>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
+      <id>openshift</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>fabric8-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>resource</goal>
+                  <goal>build</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
+</project>
+----
+
+. Create a new class in `src/main/java/com/example/`.
++
+The location of your class will vary depending on the value you set in `groupId` in your `pom.xml`. For example, for `<groupId>my.awesome.project</groupId>`, the location of the class is `src/main/java/my/awesome/project/`.
++
+.Example `src/main/java/com/example/MyApp.java`
+[source,java,options="nowrap",subs="attributes+"]
+----
+package com.example;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootApplication
+@RestController
+public class MyApp {
+
+    public static void main(String[] args) {
+        SpringApplication.run(MyApp.class, args);
+    }
+
+    @RequestMapping("/")
+    @ResponseBody
+    public Message displayMessage() {
+        return new Message();
+    }
+
+    static class Message {
+        private String content = "Greetings!";
+
+        public String getContent() {
+            return content;
+        }
+
+        public void setContent(String content) {
+            this.content = content;
+        }
+    }
+}
+----
+
+. Start your application.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn spring-boot:run
+----
+
+. Using `curl` or your browser, verify your application is running at `http://localhost:8080`.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ curl http://localhost:8080
+{"content":"Greetings!"}
+----
--- a/docs/topics/proc_deploying-an-application-to-openshift.adoc
+++ b/docs/topics/proc_deploying-an-application-to-openshift.adoc
@@ -1,0 +1,77 @@
+[#proc_deploying-an-application-to-openshift]
+= Deploying an Application to OpenShift
+
+This procedure shows you how to:
+
+* Build your application and deploy it to OpenShift using the Fabric8 Maven Plugin.
+* Use the command line to interact with your application running on OpenShift.
+
+.Prerequisites
+
+* The `oc` CLI client installed.
+* Maven installed.
+* A {SpringBoot} application.
+
+.Procedure
+
+. Log in to your OpenShift instance with the `oc` client.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ oc login ...
+----
+
+. Create a new project.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ oc new-project {project-name}
+----
+
+. In a terminal application, navigate to the directory containing your application:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ cd myApp
+----
+
+. Use Maven to start the deployment to OpenShift.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean fabric8:deploy -Popenshift
+----
++
+This command uses the Fabric8 Maven Plugin to launch the link:{link-s2i-process}[S2I process] on OpenShift and to start the pod.
+
+. Check the status of your booster and ensure your pod is running.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ oc get pods -w
+NAME                             READY     STATUS      RESTARTS   AGE
+{app-name}-1-aaaaa               1/1       Running     0          58s
+{app-name}-s2i-1-build           0/1       Completed   0          2m
+----
++
+Your `{app-name}-1-aaaaa` pod should have a status of `Running` once it is fully deployed and started.
+
+. Once your booster is deployed and started, determine its route.
++
+.Example Route Information
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ oc get routes
+NAME                 HOST/PORT                                                     PATH      SERVICES        PORT      TERMINATION
+{app-name}         {app-name}-{project-name}.{os-route-hostname}      {app-name}      8080
+----
++
+The route information of a pod gives you the base URL which you use to access it. In the example above, you would use `http://{app-name}-{project-name}.{os-route-hostname}` as the base URL to access the application.
+
+. Using `curl` or your browser, verify your application is running in OpenShift.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ curl http://{app-name}-{project-name}.{os-route-hostname}
+{"content":"Greetings!"}
+----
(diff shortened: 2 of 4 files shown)


docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+[source,xml,options="nowrap",subs="attributes+"]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.openshift.booster</groupId>
+    <artifactId>spring-boot-booster-parent</artifactId>
```
Ladicek: I don't think we want our users to depend on any kind of booster parent POM.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+    <version>1.5.10-15</version>
+  </parent>
+
+  <artifactId>my-app</artifactId>
+  <version>1.5.10-5-SNAPSHOT</version>
+  <groupId>com.example</groupId>
```
Ladicek: `<groupId>` by convention belongs right before `<artifactId>`.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.openshift.booster</groupId>
+    <artifactId>spring-boot-booster-parent</artifactId>
+    <version>1.5.10-15</version>
+  </parent>
+
+  <artifactId>my-app</artifactId>
+  <version>1.5.10-5-SNAPSHOT</version>
```
Ladicek: Is that a booster version? For an example like this, I'd go with `1.0.0-SNAPSHOT`.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+  <groupId>com.example</groupId>
+
+  <name>MyApp</name>
+  <description>My Application</description>
+
+
+
+  <dependencies>
```
Ladicek: There needs to be a `<dependencyManagement>` section right before this one, where we import the Spring Boot BOM.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+      <artifactId>cxf-spring-boot-starter-jaxrs</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.jaxrs</groupId>
+      <artifactId>jackson-jaxrs-json-provider</artifactId>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
```
Ladicek: There also needs to be the `openshift` profile, which would configure the Fabric8 Maven plugin.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
+</project>
+----
+
+. Create your class in `src/main/java/com/example/`.
++
+The location of your class will vary depending on the value you set in `groupId` in your `pom.xml`. For example, if you had `<groupId>my.awesome.project</groupId>` the location of the class would be `src/main/java/my/awesome/project/`.
```
Ladicek: It's true that by convention, the top-level package name corresponds to the Maven `groupId`, but it's not mandatory.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootApplication
+@RestController
+public class MyApp
+{
+	public static void main(String[] args) {
```
Ladicek: Please just avoid tabs.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+        SpringApplication.run(MyApp.class, args);
+    }
+
+	@RequestMapping("/")
+	@ResponseBody
+	public Message displayMessage()
+	{
```
Ladicek: The most common convention in the Java world is that the opening brace `{` belongs to the end of the previous line.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
+[#proc_creating-a-spring-boot-app]
+= Creating a {SpringBoot} Application
+
+In addition to xref:mission-http-api-spring-boot-tomcat[using a booster], you can also create new {SpringBoot} applications from scratch and deploy them to OpenShift.
```
tradej: _also_ is superfluous.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+. Using `curl` or your browser, verify your application is running at `http://localhost:8080`.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ curl http://localhost:8080
+{"content":"Greetings!"}
+----
+
+[#proc_deploy-new-app-openshift]
+== Deploying an Application to OpenShift
```
tradej: Should be a separate file, because modules are always one module per file.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
+[#proc_creating-a-spring-boot-app]
```
tradej: The `proc_` prefix is not used in anchor IDs.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+= Creating a {SpringBoot} Application
+
+In addition to xref:mission-http-api-spring-boot-tomcat[using a booster], you can also create new {SpringBoot} applications from scratch and deploy them to OpenShift.
+
+== Creating an Application
```
tradej: Should be a separate module.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
+</project>
+----
+
+. Create your class in `src/main/java/com/example/`.
```
tradej: Replace _your_ with _a new._

docs/topics/proc_creating-spring-boot-app.adoc
tradej: Replace _if you had_ with _for._ There should be a comma before _the location._

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
+[#proc_creating-a-spring-boot-app]
+= Creating a {SpringBoot} Application
+
+In addition to xref:mission-http-api-spring-boot-tomcat[using a booster], you can also create new {SpringBoot} applications from scratch and deploy them to OpenShift.
+
+== Creating an Application
+
+.Prerequisites
+* `mvn` installed.
```
tradej: I would prefer _Maven_ instead of `mvn`. The latter is just the binary, whereas you need the complete build system.

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+.Procedure
+
+. Create the application folder.
```
tradej: _Folder_ is Windows terminology. Please use _directory._

docs/topics/proc_creating-spring-boot-app.adoc
inoxx03: I agree this is too specific to be an example:) I will change it to `1.0.0-SNAPSHOT`

docs/topics/proc_creating-spring-boot-app.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+. Using `curl` or your browser, verify your application is running at `http://localhost:8080`.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ curl http://localhost:8080
+{"content":"Greetings!"}
+----
+
+[#proc_deploy-new-app-openshift]
```
tradej: The anchor ID should be "deploying-an-application-to-openshift".

docs/topics/proc_creating-spring-boot-app.adoc
inoxx03: Thank you. I've been intending to do that:)

docs/topics/proc_creating-spring-boot-app.adoc
inoxx03: Agreed.  Besides, i think we use the full form in other similar instances in the docs

inoxx03:
@Ladicek  @tradej Thank you for the feedback. I have incorporated your suggestions, Please take a look and let me know your thoughts.

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,159 @@
...
+      <id>openshift</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>fabric8-maven-plugin</artifactId>
+            <configuration>
+              <systemPropertyVariables>
+                <app.name>${project.artifactId}</app.name>
+              </systemPropertyVariables>
```
Ladicek: I'm not sure you need to do this here. You can probably remove the entire `<configuration>` block. But what you do need to add is this:

```xml
                <executions>
                    <execution>
                        <goals>
                            <goal>resource</goal>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
```

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,159 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootApplication
+@RestController
+public class MyApp
+{
+	public static void main(String[] args) {
+     SpringApplication.run(MyApp.class, args);
```
Ladicek: Still has tabs and weird indentation. Also `class MyApp {`, not `class MyApp <line break> {`.

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,159 @@
...
+              <systemPropertyVariables>
+                <app.name>${project.artifactId}</app.name>
+              </systemPropertyVariables>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>openshift-it</id>
```
Ladicek: This is our convention for running OpenShift tests. I think it's a pretty good convention, but you don't have any OpenShift tests here, so not sure if it isn't confusing.

inoxx03: @Ladicek Oh, right. The example here does not contain any tests, so this could be somewhat confusing. Let me remove that:)

docs/topics/proc_creating-an-application.adoc
inoxx03: Thank you. According to the FMP docs , the `<configuration>` block is optional, so I'm taking it out of this example. Also, I'm adding the `<execution>` block.

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,146 @@
...
+                <goals>
+                  <goal>resource</goal>
+                  <goal>build</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>s
```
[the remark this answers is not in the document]
inoxx03: Whoah, how did that get there? Thanks for catching it:)

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,145 @@
...
+----
+package com.example;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@SpringBootApplication
```
inoxx03: @Ladicek Thank you for the feedback. I've removed tabs from the code example and set the indentation to 4 spaces (AFAIK, that is the convention for Java indentation). WDYT?

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,145 @@
...
+        private String content = "Greetings!";
+
+            public String getContent() {
+                return content;
+            }
+
+            public void setContent(String content) {
+                this.content = content;
+            }
+    }
```
Ladicek: Indentation is wrong in this class. Also, it really should be a `static class`.

docs/topics/proc_creating-an-application.adoc
Ladicek: Agree, it's much easier to read now. I added one more comment about the nested class `Message` though :-)

docs/topics/proc_creating-an-application.adoc
```diff
@@ -0,0 +1,146 @@
+[#creating-an-application]
+= Creating an application
+
```
tradej: I think this chapter could use an introduction if we wanted to it really properly, but I won't hold this review on it.

docs/topics/proc_deploying-an-application-to-openshift.adoc
```diff
@@ -0,0 +1,64 @@
+[#proc_deploying-an-application-to-openshift]
+= Deploying an Application to OpenShift
+
```
tradej: Same here, an introduction would be nice (what application, what sort of deployment this is, etc.).

docs/topics/proc_deploying-an-application-to-openshift.adoc
```diff
@@ -0,0 +1,64 @@
...
+.Prerequisites
+
+* The `oc` CLI client installed.
+* Maven installed.
```
tradej: You also need the application.

docs/topics/proc_deploying-an-application-to-openshift.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+.Procedure
```
tradej: It may be good to say this happens in the directory with the application, but I'll leave that to you.

docs/topics/proc_deploying-an-application-to-openshift.adoc
inoxx03: Yeah, that makes sense:) thank you for pointing that out.

docs/topics/proc_deploying-an-application-to-openshift.adoc
inoxx03: Thank you, that is important information. Let me add it there:)

docs/topics/proc_creating-an-application.adoc
inoxx03: Thank you, I think including an introduction here is beneficial. I'll add one here as well as in the deployment section.

docs/topics/proc_creating-an-application.adoc
inoxx03: @Ladicek  Thank you:)  I've de-indented the `setContent` and `getContent` methods. The indentation  seem to make sense now:) Actually, I would like to know why `static class Message` is the proper definition. Let's discuss this via email :)

docs/topics/proc_deploying-an-application-to-openshift.adoc
```diff
@@ -0,0 +1,78 @@
+[#proc_deploying-an-application-to-openshift]
+= Deploying an Application to OpenShift
+
+This procedure shows you how to:
+
+* Build your {SpringBoot} application from source and create an application image using the S2I build process.
```
Ladicek: It technically uses S2I (FMP uses S2I binary source builds), but from user's point of view, S2I is something completely different. I wouldn't mention S2I here.

rhoads-zach:
Thank you @inoxx03 for the updates. Merging.

PR #1065, opened 2018-03-28
fixes RHOARDOC-1216

inoxx03:
Reusability changes for the App from scratch content.
+ it makes the starter app development assembly shareable by Vert.x  and Spring Boot.
+ now with comments providing parameterization reference
+ still uses way too many parameters, but works :)

### DIFF
--- a/docs/topics/resources/vertx-starter-example-pom.xml
+++ b/docs/topics/resources/vertx-starter-example-pom.xml
@@ -12,8 +12,8 @@
   <description>Example application using RHOAR Vert.x</description>
 
   <properties>
-    <vertx.version>3.4.2.redhat-009</vertx.version>
-    <vertx-maven-plugin.version>1.0.13</vertx-maven-plugin.version>
+    <vertx.version>{VertXVersion}</vertx.version>
+    <vertx-maven-plugin.version>{VertXMvnPluginVersion}</vertx-maven-plugin.version>
     <vertx.verticle>io.openshift.booster.MyApplication</vertx.verticle>
   </properties>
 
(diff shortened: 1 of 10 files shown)


inoxx03:
@rhoads-zach @tradej Please let me know what you think.

docs/vertx-runtime/master.adoc
```diff
@@ -255,11 +255,21 @@ include::topics/secured-mission-resources.adoc[leveloffset=+3]
 //Minimum Viable App
 //Add-ons
 //Building your app
+// TODO: make content properly reusable
+//creating an app from scratch
```
Ladicek: Does the TODO still make sense? :-)

inoxx03: @Ladicek Oh, right. That one is no longer necessary. Guess I should remove that :)

docs/topics/resources/vertx-starter-example-pom.xml
```diff
@@ -12,8 +12,8 @@
   <description>Example application using RHOAR Vert.x</description>
 
   <properties>
-    <vertx.version>3.4.2.redhat-009</vertx.version>
-    <vertx-maven-plugin.version>1.0.13</vertx-maven-plugin.version>
+    <vertx.version>{VertXVersion}</vertx.version>
+    <vertx-maven-plugin.version>{VertXMvnPluginVersion}</vertx-maven-plugin.version>
```
Ladicek: I see `VertXVersion` defined in `document-attributes.adoc`, but where is `VertXMvnPluginVersion` defined?

inoxx03: @Ladicek Thank you for calling this out. I've actually noticed that myself a while ago and added the `VertxMvnPluginVersion` variable to`document attributes`, which store the VMP version.

Ladicek: Cool, thanks!

PR #1066, opened 2018-03-29
Fixed missing version in WFS App from scratch example

tradej:
The `:version:` attribute setting needed to precede the inclusion, 'tis all.

Fixes RHOARDOC-1213


inoxx03:
merged: 75e1c49b0e7e71b89cc68b72ac4372037c6d1088

PR #1067, opened 2018-03-29
updated spring boot bom version. Fixes RHAORDOC-1210

rhoads-zach:


rhoads-zach:
this is unneeded based on #1066

Ladicek:
Thanks!

PR #1070, opened 2018-04-03
added callout to BOM. Fixes RHOARDOC-1217

rhoads-zach:


docs/spring-boot-runtime/master.adoc
```diff
@@ -240,6 +240,8 @@ include::topics/proc_running-the-secured-booster-integration-tests-on-sb-vx.adoc
 include::topics/secured-mission-resources.adoc[leveloffset=+3]
 
 == Application Development
+The recommended approach for specifying and using supported and tested maven artifacts in a {SpringBoot} application is to use the OpenShift Application Runtimes {SpringBoot} BOM.
```
Ladicek: Do we have a variable somewhere for the "OpenShift Application Runtimes" text? Not sure if it would fit into the present phrasing, but I guess care must be taken here.

inoxx03: @Ladicek, we do have a `ProductName` variable in `document-attributes.adoc`, which resolves to `Red Hat OpenShift Application Runtimes` in the _downstream_ docs only.
In the upstream, it resolves to `Application Development on OpenShift` to avoid branding the upstream docs.
I don't think _Application Development on OpenShift_ would fit inn the current phrasing, so I'd be inclined to leave the paragraph as is for the time being. We can always define a new attribute later:)

Ladicek: Sure, there's probably some reason why we don't use the phrase "Red Hat OpenShift Application Runtimes" in the upstream docs, and "OpenShift Application Runtimes" is suspiciously similar to that :-)

rhoads-zach:
merged 73311b6e0f32aa14f018d886cb8d1ac95b096140

PR #1072, opened 2018-04-03
updated package.json examples. Fixes RHOARDOC-1220

rhoads-zach:


docs/topics/templates/document-attributes.adoc
```diff
@@ -97,6 +97,8 @@
 :VertXVersion: 3.4.2.redhat-009
 :VertXMvnPluginVersion: 1.0.13
 
+:nodeshiftNodeVersion: --nodeVersion=8.x
+//:nodeshiftNodeVersion: --dockerImage=registry.access.redhat.com/rhoar-nodejs/nodejs-8
```
Ladicek: Is the intent here to switch to the latter downstream? We _so_ need to figure out the relationship of our upstream docs to our upstream projects, because we're not really consistent.

rhoads-zach: @Ladicek yes. I can remove the commented line and add it downstream only if you like.

Ladicek: In fact, I'd prefer if upstream docs just documented downstream product :-) I know that's what we do on some other places (e.g. all of WildFly Swarm).

rhoads-zach: thank you @Ladicek updated PR per feedback.

PR #1074, opened 2018-04-04
updated redirect rules. Fixes RHOARDOC-1224

rhoads-zach:


rhoads-zach:
@gastaldi @tradej @inoxx03 mind taking a look at this?

gastaldi:
@rhoads-zach will try it locally later, but do you really need to verify each mission in if statements? Why not use the value in the argument directly?

tradej:
I don't see anything explicitly wrong with it, but the nginx [FAQ](http://nginx.org/en/docs/faq/variables_in_config.html) says variables are costly and should be avoided, so this is probably not what they had in mind.

rhoads-zach:
i can update to remove the vars, but we have an inconsistency in naming. The following two missions uses LHS in the launcher and RHS in the docs:
* `rest-http` -> `http-api`
* `rest-http-secured` -> `secured`

I can modify this in the docs in this PR to make it consistent and skip setting the vars. @tradej @inoxx03 @Ladicek any objections?

Ladicek:
No objection to changing the anchors.

rhoads-zach:
@gastaldi @inoxx03 @tradej @Ladicek i changed the anchors, which allowed me to clean up the redirects.

We still have some attributes that use the old scheme, but they do not affect the anchors.  I'll clean those up in a follow up PR.

inoxx03:
merging...

inoxx03:
merged: 58418be72bb8b4a3326c10eba9adb74bf6aef300

PR #1075, opened 2018-04-04
Deploying an existing WFS app to OpenShift (RHOARDOC-1195)

tradej:
Resolves RHOARDOC-1195

Likely reusable by other Maven-based runtimes with minor modifications.

### DIFF
--- a/docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
+++ b/docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
@@ -13,7 +13,7 @@ You can easily deploy your existing application to OpenShift using the Fabric8 M
 . Add the following profile to the `pom.xml` file in the root directory of your application:
 +
 --
-[source,xml,options="nowrap"]
+[source,xml,options="nowrap",subs="attributes+"]
 ----
 <profile>
   <id>openshift</id>
@@ -22,6 +22,7 @@ You can easily deploy your existing application to OpenShift using the Fabric8 M
       <plugin>
         <groupId>io.fabric8</groupId>
         <artifactId>fabric8-maven-plugin</artifactId>
+        <version>{Fabric8MavenPluginVersion}</version>
         <executions>
           <execution>
             <goals>
--- a/docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
+++ b/docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
@@ -1,0 +1,59 @@
+
+[id='deploying-an-existing-wildfly-swarm-application-to-openshift]
+= Deploying an existing {WildFlySwarm} application to OpenShift
+
+You can easily deploy your existing application to OpenShift using the Fabric8 Maven plugin.
+
+.Prerequisites
+
+* A {WildFlySwarm}{ndash}based application
+
+.Procedure
+
+. Add the following profile to the `pom.xml` file in the root directory of your application:
++
+--
+[source,xml,options="nowrap"]
+----
+<profile>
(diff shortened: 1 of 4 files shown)


docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
```diff
@@ -0,0 +1,60 @@
...
+        <artifactId>fabric8-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>resource</goal>
+              <goal>build</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
```
tradej: Not sure what this section does or if it should be here.

docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
```diff
@@ -0,0 +1,60 @@
...
+              <wildfly-swarm-health-check>
+                <path>/</path>
+              </wildfly-swarm-health-check>
+            </config>
+          </enricher>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+  <activation />
```
tradej: Not sure what this does.

docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
Ladicek: That's something we'll need to verify and possibly update in the chapter on creating new app from scratch. I vaguely recollect that excluding the `webapp` generator used to be necessary for Swarm, but not sure if that still applies.

docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
Ladicek: This line can be removed.

docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
```diff
@@ -0,0 +1,60 @@
...
+--
+[source,xml,options="nowrap"]
+----
+<profile>
+  <id>openshift</id>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>io.fabric8</groupId>
+        <artifactId>fabric8-maven-plugin</artifactId>
```
Ladicek: Should probably show that specifying `<version>` is necessary.

tradej: Done.

docs/topics/proc_deploying-an-existing-wildfly-swarm-application-to-openshift.adoc
```diff
@@ -0,0 +1,60 @@
...
+      </plugin>
+    </plugins>
+  </build>
+</profile>
+----
+
+In this profile, the Fabric8 Maven plugin is invoked for building and deploying the application to OpenShift.
+--
+
+. Deploy the application to OpenShift according to instructions in xref:deploying-an-application-to-openshift[].
```
Ladicek: Should this line in fact be an `include::topics/...something...`?

inoxx03: @Ladicek it's true that including the module would also work here, but in this specific instance, we're cross-referencing a section that immediately precedes this content in the current guide structure. including the model again would make the section a bit redundant. An alternative would be to flip these 2 modules so that the deployment instructions follow rather than precede the section, but IMHO it is not a big issue. I won't hold this PR on it:)

PR #1079, opened 2018-04-06
fixes RHOADOC-1230 for Vertx

inoxx03:
@Ladicek Added the <repositories> and <pluginRepositories> blocks for the Maven GA repo. Please let me know if this is OK.

### DIFF
--- a/docs/topics/resources/vert-x/vertx-starter-example-pom.xml
+++ b/docs/topics/resources/vert-x/vertx-starter-example-pom.xml
@@ -40,7 +40,7 @@
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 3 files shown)


Ladicek:
LGTM, but we need it for Spring Boot and WildFly Swarm as well :-)

docs/topics/resources/vert-x/vertx-starter-example-pom.xml
```diff
@@ -40,6 +40,23 @@
     </dependency>
   </dependencies>
 
+<!-- Specify the repositories containing RHOAR artifacts -->
```
Ladicek: Maybe indent this line like the other lines -- I guess it's visible in the rendered docs.

PR #1061, opened 2018-03-28
add caching mission

rhoads-zach:

PR #1083, opened 2018-04-09
fixes RHOARDOC-1227

inoxx03:
@rhoads-zach I've added the `integrationTestSection` variable containing a `NOTE: ` to the launcher generated `README.adoc` files. The note contains the command to run the integration tests (for boosters that contain them).
NOTE: I have not yet done so for the `osio-*-README.adoc`, but will do so, if we agree to apply this change.

The idea behind this is that if we include the test instructions in the README, users won't have to look at the detailed docs when getting started with the booster.

I haven't been able to apply formatting to  a section stored inside a variable (`\n` does not work for line breaks), so this solution circumvents the need for formatting by keeping the minimum instructions inside a `NOTE` admonition block. (Admittedly not the squeaky-cleanest way of doing so, but IMHO gets the point across)

@rhoads-zach @tradej Please let me know what you think.

### DIFF
--- a/docs/topics/readme/cd-circuit-breaker-spring-boot.properties
+++ b/docs/topics/readme/cd-circuit-breaker-spring-boot.properties
@@ -5,5 +5,6 @@ fileLocation=src/main/resources/static/index.html
 greeting-app-name=springboot-cb-greeting
 name-app-name=springboot-cb-name
 gitCMD=git clone git@github.com:${loggedUser}/${targetRepository}
-cicdSection=IMPORTANT: As part of the process of creating this booster, developers.redhat.com/launch or the Fabric8 Launcher tool set up a project with a CI/CD deployment of this booster. You can see the status of this deployment in your Single-node OpenShift Cluster or OpenShift Online Web Console. 
+cicdSection=IMPORTANT: As part of the process of creating this booster, developers.redhat.com/launch or the Fabric8 Launcher tool set up a project with a CI/CD deployment of this booster. You can see the status of this deployment in your Single-node OpenShift Cluster or OpenShift Online Web Console.
 prerequisite=IMPORTANT: This booster requires Java 8 JDK or greater and Maven 3.3.x or greater.
+integrationTestSection=NOTE: Run the set of integration tests included with this booster using `mvn clean verify -Popenshift,openshift-it`.
--- a/docs/topics/readme/osio-rest-http-secured-wildfly-swarm.properties
+++ b/docs/topics/readme/osio-rest-http-secured-wildfly-swarm.properties
@@ -1,2 +1,3 @@
 guideURL=http://appdev.openshift.io/docs/wf-swarm-runtime.html
 fileLocation=src/main/webapp/index.html
+integrationTestSection=To run the integration test on the Secured HTTP API booster follow the steps in the link:https://appdev.openshift.io/docs/wf-swarm-runtime.html#running-the-secured-booster-integration-tests[Secured Booster documentation].
--- a/docs/topics/readme/rest-http-secured-README.adoc
+++ b/docs/topics/readme/rest-http-secured-README.adoc
@@ -26,6 +26,7 @@ $ oc create -f service.sso.yaml
 $ ${localRunCMD}
 ----
 
+${integrationTestSection}
 
 == Interacting with the Booster on a Single-node OpenShift Cluster
 
--- a/docs/topics/readme/zip-health-check-nodejs.properties
+++ b/docs/topics/readme/zip-health-check-nodejs.properties
@@ -7,3 +7,4 @@ gitCMD=unzip ${artifactId}.zip
 targetRepository=${artifactId}
 cicdSection=
 prerequisite=IMPORTANT: This booster requires Node.js 8.x or greater and `npm` 5 or greater.
+integrationTestSection=
--- a/docs/topics/readme/cd-circuit-breaker-vert.x.properties
+++ b/docs/topics/readme/cd-circuit-breaker-vert.x.properties
@@ -5,5 +5,6 @@ fileLocation=src/main/resources/webroot/index.html
 greeting-app-name=greeting-service
 name-app-name=name-service
 gitCMD=git clone git@github.com:${loggedUser}/${targetRepository}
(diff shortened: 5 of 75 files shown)


docs/topics/readme/cd-circuit-breaker-spring-boot.properties
```diff
@@ -5,5 +5,6 @@ fileLocation=src/main/resources/static/index.html
 greeting-app-name=springboot-cb-greeting
 name-app-name=springboot-cb-name
 gitCMD=git clone git@github.com:${loggedUser}/${targetRepository}
-cicdSection=IMPORTANT: As part of the process of creating this booster, developers.redhat.com/launch or the Fabric8 Launcher tool set up a project with a CI/CD deployment of this booster. You can see the status of this deployment in your Single-node OpenShift Cluster or OpenShift Online Web Console. 
-prerequisite=IMPORTANT: This booster requires Java 8 JDK or greater and Maven 3.3.x or greater.
\ No newline at end of file
+cicdSection=IMPORTANT: As part of the process of creating this booster, developers.redhat.com/launch or the Fabric8 Launcher tool set up a project with a CI/CD deployment of this booster. You can see the status of this deployment in your Single-node OpenShift Cluster or OpenShift Online Web Console.
+prerequisite=IMPORTANT: This booster requires Java 8 JDK or greater and Maven 3.3.x or greater.
+integrationTestSection=NOTE: Run the set of integration tests included with this booster using `$ mvn clean verify -Popenshift,openshift-it`.
```
Ladicek: I understand that the usage of `$` is meant to signify a shell prompt, but I think that only makes sense in larger code blocks. When you include a command inline into a sentence, I think including the prompt marker does more harm than good.

docs/topics/readme/rest-http-secured-README.adoc
```diff
@@ -26,6 +26,7 @@ $ oc create -f service.sso.yaml
 $ ${localRunCMD}
 ----
 
+${integrationTestSection}
```
Ladicek: For the Secured boosters, this variable is not defined. Also it's unfortunately not as simple as for the other boosters, so I'd rather comment this out for now.

(There's unfortunately a difference between boosters here. The Vert.x integration tests deploy the SSO server as part of the test, while Swarm and Spring Boot tests don't.)

Ladicek:
Otherwise LGTM.

docs/topics/readme/cd-circuit-breaker-spring-boot.properties
tradej: I concur, we typically include the shell prompt only in code blocks (as opposed to inline code) because in sentences, you can easily say "as a regular user" or "as root" and it's natually part of the information.

docs/topics/readme/rest-http-secured-README.adoc
Ladicek: This commet is still valid.

Also, should probably add the same info to https://github.com/fabric8-launcher/launcher-documentation/pull/1061 or plan to do the same after.

inoxx03: @Ladicek Thank you. Given the relatively subtle but important differences in the way integration tests are executed against the Secured booster on different runtimes, I propose we include a link to the integration test section for the respective booster rather than citing specific steps and/or command examples in the README.

docs/topics/readme/cd-circuit-breaker-spring-boot.properties
inoxx03: @Ladicek @tradej Thank you for the feedback. I agree, including the shell prompt does not really make sense for an inline command example like this. Removing...

docs/topics/readme/rest-http-secured-README.adoc
Ladicek: Works for me, thanks!

docs/topics/readme/osio-rest-http-secured-wildfly-swarm.properties
```diff
@@ -1,2 +1,4 @@
 guideURL=http://appdev.openshift.io/docs/wf-swarm-runtime.html
-fileLocation=src/main/webapp/index.html
\ No newline at end of file
+fileLocation=src/main/webapp/index.html
+integrationTestSection=NOTE: Run the set of integration tests included with this booster using `mvn clean verify -Popenshift,openshift-it`.
+integrationTestSection=To run the integration test on the Secured HTTP API booster follow the steps in the link:https://appdev.openshift.io/docs/wf-swarm-runtime.html#running-the-secured-booster-integration-tests[Secured Booster documentation].
```
Ladicek: Only the 2nd `integrationTestSection` is correct.

docs/topics/readme/rest-http-secured-README.adoc
```diff
@@ -127,4 +128,6 @@ $ > GET /api/greeting HTTP/1.1
 > Accept: */*
 > Authorization: Bearer <TOKEN>
 ----
-====
\ No newline at end of file
+====
+
+$(integrationTestSection)
```
Ladicek: Why add it twice in the same README?

docs/topics/readme/zip-health-check-nodejs.properties
```diff
@@ -6,4 +6,6 @@ app-name=booster-health-check-nodejs
 gitCMD=unzip ${artifactId}.zip
 targetRepository=${artifactId}
 cicdSection=
-prerequisite=IMPORTANT: This booster requires Node.js 8.x or greater and `npm` 5 or greater.
\ No newline at end of file
+prerequisite=IMPORTANT: This booster requires Node.js 8.x or greater and `npm` 5 or greater.
+integrationTestSection=
+integrationTestSection=NOTE: Run the set of integration tests included with this booster using `mvn clean verify -Popenshift,openshift-it`.
```
Ladicek: Only the 1st `integrationTestSection` line is correct.

Ladicek:
OK otherwise :-)

docs/topics/readme/zip-health-check-nodejs.properties
inoxx03: Thanks, removed the incorrect one.

docs/topics/readme/rest-http-secured-README.adoc
inoxx03: Honestly, It was a brain fart on my part:) Totally unintended. Removing the extra one at the bottom of the file. Thank you for spotting this.

PR #1087, opened 2018-04-11
fixes RHOARDOC-1239

inoxx03:
According to information form @cescoffier, the `<exclude>f8-maven-scm</exclude>` is not required and the `<vertx-health-check>` should be replaced by the `<vertx.health>` property. 

@Ladicek Please let me know what you think

Additional actions: Added explanatory comments to the Vert.x POM example.

### DIFF
--- a/docs/spring-boot-runtime/master.adoc
+++ b/docs/spring-boot-runtime/master.adoc
@@ -242,18 +242,13 @@ include::topics/secured-mission-resources.adoc[leveloffset=+3]
 == Application Development
 The recommended approach for specifying and using supported and tested maven artifacts in a {SpringBoot} application is to use the OpenShift Application Runtimes {SpringBoot} BOM.
 
+//creating an app from scratch
 // workaround to resolve ID conflicts:
 :context: spring-boot
-//creating an app from scratch
-include::topics/assembly_creating-a-basic-spring-boot-application.adoc[leveloffset=+2]
-
-:parameter-maven-command: $ mvn spring-boot:run
-:parameter-runtime-name: {SpringBoot}
 :parameter-runtime: spring-boot
+include::topics/assembly_creating-a-basic-spring-boot-application.adoc[leveloffset=+2]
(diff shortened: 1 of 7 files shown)


Ladicek:
I don't think "the <vertx-health-check> should be replaced by the <vertx.health> property", it _can_ be. I'd personally prefer keeping it in the FMP configuration, because it's just FMP who uses this value to automatically generate a health check.

Ladicek:
Actually, the property name _isn't_ `vertx.health`, it's `vertx.health.path`. And there's yet another property for readiness check; this one is only used for liveness. (C.f. https://github.com/fabric8io/fabric8-maven-plugin/blob/master/enricher/fabric8/src/main/java/io/fabric8/maven/enricher/fabric8/VertxHealthCheckEnricher.java)

I'd personally just got rid of the health checks in these chapters...

inoxx03:
@Ladicek Thank you for the feedback and the explanation around the health check conf:) If the health check configuration is not required, let's just leave it out of the basic app example altogether.

Ladicek:
Adding health checks is a best practice, but not a hard requirement. I think we could just add a link to the documentation of the health check booster and be done :-)

inoxx03:
@Ladicek Yeah, that's a good point:) I'll just add a link under the _Additional information_ section. Thank you:)

inoxx03:
@Ladicek thank you for the feedback.
Actions: 
* updated and amended PR
* incorporated feedback
* added link to health check booster as additional info (under _Creating an Application_)

@rhoads-zach Please feel free to merge if this looks OK to you.

Thank you :)

Ladicek:
Nice, thanks!

@inoxx03 The paragraph with recommendation to add health check and link to the Health Check booster will be available in all 4 runtime guides, correct?

inoxx03:
@Ladicek Yes, the paragraph will be available for all 4 runtimes (SB, VX, WFS, & Node). I've set the link up with a `parameter-runtime` so that it always points to the Health Check booster for the correct runtime.

PR #1089, opened 2018-04-11
Added appendix on deploying booster using the template. Fixes RHOARDO…

rhoads-zach:
…C-1196

### DIFF
--- a/docs/nodejs-runtime/master.adoc
+++ b/docs/nodejs-runtime/master.adoc
@@ -239,6 +239,9 @@ include::topics/appendix-nodeshift.adoc[leveloffset=+1]
 include::topics/http-api-nodejs-package-json.adoc[leveloffset=+1]
 
 [appendix]
+include::topics/proc_deploying-a-booster-using-openshift-templates.adoc[leveloffset=+1]
+
+[appendix]
 include::topics/nodejs-dev-guide-additional-resources.adoc[leveloffset=+1]
 
 [appendix]
--- a/docs/spring-boot-runtime/master.adoc
+++ b/docs/spring-boot-runtime/master.adoc
@@ -313,6 +313,9 @@ include::topics/appendix-using-war-files-spring-boot.adoc[leveloffset=+1]
 //include::topics/http-api-spring-boot-pom.adoc[leveloffset=+1]
 
 [appendix]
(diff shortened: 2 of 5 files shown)


Ladicek:
1. This will only work for the simplest boosters we have. Some boosters have more than one `application.yaml` file, and we need to deploy all of them to create the functional application. For that, the `oc new-app` approach doesn't seem like the best fit. I'd instead advocate for `oc process -f application.yaml | oc apply -f -`.
1. Some boosters also require other parameters beyond `SOURCE_REPOSITORY_URL`. In the automated tests, we set `SOURCE_REPOSITORY_REF` and `SOURCE_REPOSITORY_DIR`. And the Launcher supports even more; fortunately, these 3 seem to be enough for now.
1. We certainly shouldn't advocate changing the files directly. Both `oc process` and `oc new-app` commands have the `-p` option to supply the parameter values.
1. Some boosters also require deploying more files than just `application.yaml` -- such as the ConfigMap booster (`resource.configmap.yalm`) or CRUD (`service.yaml` for PostgreSQL). Content of those files could be inlined into a single template, but that isn't the case now. Which brings me to...
1. I agree that we definitely have to describe the S2I deployment, but for that, we first need proper templates. What we have now in the `.openshiftio` directories is only meant to be used by the Launcher, is maintained semi-automatically (and contains a lot of crap), and my impression is that everyone treats it as a burden, not as an opportunity. We need to change that attitude first.
1. Also, `.openshiftio` is an internal directory. I don't think we should depend on its content in userfacing procedures. Which finally leads me to...
1. When creating the `.zip` or Git repo, Launcher removes the `.openshiftio` directories. So we can't depend on it anyway.

Sorry to say that, but I don't think we're in a position where we could, with confidence, provide instructions for S2I deployment.

quintesse:
@Ladicek what you propose would certainly be an improvement. But it would also mean that each runtime team would have to manually maintain (and test) that `deploy.yaml` file. THe reason why the Launcher uses a much more complex system is because it's partly generated by the mvn fabric8 plugin. At the time that seemed the best way so the runtime teams wouldn't have to know too much about what was needed to run on OpenShift. With what you're suggesting they _will_ need to know. (Not saying that as a bad thing, but just as something that will need to be taken into consideration)

Ladicek:
@quintesse You're absolutely right and I still think that we _must_ understand our OS. Pun intended :-)

PR #1090, opened 2018-04-12
added more context on pod name. Fixes RHOARDOC-4

rhoads-zach:


inoxx03:
merged: c53518db52f8a5f85068a11bc87d44cd4ea89b97

Ladicek:
I don't see anything factually wrong with this, but I'd be wary of spending our cognitive load budget with minutiae like this. I believe everyone would intuitively understand how the pod names work.

PR #1094, opened 2018-04-16
updated sso mission instructions based on feedback. Fixes RHOARDOC-1197

rhoads-zach:


docs/topics/proc_interacting-with-the-secured-booster-on-the-command-line.adoc
```diff
@@ -40,11 +40,7 @@ $ curl -sk -X POST https://<SSO_AUTH_SERVER_URL>/auth/realms/$REALM/protocol/ope
   -d client_secret=$SECRET
 ----
 +
-NOTE: The `-sk` option tells curl to ignore failures resulting from self-signed certificates. Do not use this option in a production environment.
-+
-On macOS, you must have `curl` version `7.56.1` or greater installed. It must also be built with OpenSSL.
-
-** Extract the access token information, for example using the `jq` tool:
+Optionally, you can extract the access token information, for example using the `jq` tool:
```
Ladicek: I don't think that extracting the token is optional. You have to do that in order to invoke the service. What can be considered optional is the usage of the `jq` tool.

Ladicek: But what RHOARDOC-1197 actually seems to ask for is a _single_ command that contains both `curl ... | jq ...`. I would personally agree that it's much better than splitting the command to two, but I guess you have split it for a reason (educative probably?).

docs/topics/readme/rest-http-secured-README.adoc
```diff
@@ -95,16 +95,16 @@ $ curl -sk -X POST https://<SSO_AUTH_SERVER_URL>/auth/realms/$REALM/protocol/ope
   -d client_secret=$SECRET
 ----
 +
-NOTE: The `-sk` option tells curl to ignore failures resulting from self-signed certificates. Do not use this option in a production environment.
-
-. Extract the access token information using the `jq` tool:
+Optionally, you can extract the access token information, for example using the `jq` tool:
```
Ladicek: Same as above, extracting the token is mandatory; what's optional is using the `jq` tool.

docs/topics/secured-mission-intro.adoc
```diff
@@ -10,7 +10,7 @@ The SSO booster project contains:
 = Standing up {RHSSO}
 The `service.sso.yaml` file contains all OpenShift configuration items to stand up a pre-configured {RHSSO} server. The SSO server configuration has been simplified for the sake of this exercise and does provide an out-of-the-box configuration, with pre-configured users and security settings.
 
-IMPORTANT: It is not recommended to use this SSO configuration in production. Specifically, the simplifications made to the booster security configuration impact the ability to use it in a production environment.
+WARNING: The `service.sso.yaml` file contains very long lines, and some text editors, such as link:https://wiki.gnome.org/Apps/Gedit[gedit], may have issues reading this file. It is also not recommended to use this SSO configuration in production. Specifically, the simplifications made to the booster security configuration impact the ability to use it in a production environment.
```
Ladicek: So here we conflate two warnings, one that has a very little severity (`gedit` is buggy) and one that has extremely high severity (don't use this config in production). I think this is a bad idea. Especially if the high-severity issue is mentioned _after_ the low-severity issue.

Ladicek:
This also doesn't address the first item from RHOARDOC-1197, namely that there are two SSO-related routes in the output of `oc get routes`. Looking at https://appdev.prod-preview.openshift.io/docs/spring-boot-runtime.html#getting-the-secured-booster-api-endpoint, it's not only chapter 3.6.4, but also 3.6.6: the _Getting the Secured Booster API Endpoint_ procedure doesn't really say which route is the right one. Same in the booster README.

rhoads-zach:
@Ladicek updated based on feedback and address point 1

PR #1102, opened 2018-04-24
bump launcher template version to v18

inoxx03:


rhoads-zach:
@inoxx03 LGTM

Ladicek:
Weren't there like 3 places where you need to update the template version? Or where the other 2 in the Launcher install script, which is now gone?

rhoads-zach:
@Ladicek the other entries were for the script, so yes those are now gone

PR #1110, opened 2018-04-26
fixed bad anchors. Fixes RHOARDOC-1279

rhoads-zach:


docs/topics/assembly_creating-a-basic-vertx-application.adoc
```diff
@@ -1,5 +1,5 @@
 
-[id="creating-a-basic-spring-boot-application_{context}"]
+[id="creating-a-basic-vertx-application_{context}"]
```
Ladicek: Just wondering if the `{context}` bit doesn't serve the same purpose?

rhoads-zach:
@Ladicek good call, fixed.

PR #1116, opened 2018-05-04
Updated the "running integration tests" part in Vert.x secured booste…

spisiakm:
…r guide, fixes https://issues.jboss.org/browse/RHOARDOC-1281


docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
```diff
@@ -0,0 +1,37 @@
+// This is a parameterized module. Parameters used:
+//
+//   parameter-runtime: runtime name in the module ID
+//
+// Rationale: This parameter displays the correct runtime name (`sb` or `vx`) in the module ID that is visible when linking to this section.
```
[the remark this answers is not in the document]
spisiakm: You're right. This was just copied over, to be deleted.

Is the parameter `parameter-runtime` used at all though? In `proc_running-the-secured-booster-integration-tests-on-sb-vx`, it looks like the lines where the parameter is used are commented out.

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
```diff
@@ -0,0 +1,37 @@
...
+. Execute the integration tests:
++
+[source,bash,options="nowrap",subs="attributes+"]
+--
+mvn clean verify -Popenshift,openshift-it
+--
+
+You can also skip the SSO server deployment in case you have deployed a SSO server instance beforehand, e.g. by executing `oc create -f service.sso.yaml`.
+In that case, you can set an additional maven property `skip.sso.init` to `true`, similar to example below:
+[source,bash,options="nowrap",subs="attributes+"]
```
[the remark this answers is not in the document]
spisiakm: Don't know, when I built it, it looked alright.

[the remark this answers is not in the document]
Ladicek: OK, I dunno either, I'm no Asciidoc expert, I only know I've seen `[source]...` blocks surrounded by empty lines everywhere :-)

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+. In a terminal application, navigate to the directory with your project.
+. Execute the integration tests:
++
+[source,bash,options="nowrap",subs="attributes+"]
+--
+mvn clean verify -Popenshift,openshift-it
+--
+
+You can also skip the SSO server deployment in case you have deployed a SSO server instance beforehand, e.g. by executing `oc create -f service.sso.yaml`.
```
Ladicek: This sentence could probably use some rewording. I know that "e.g. by executing `oc create -f service.sso.yaml`" is _meant to_ bind to "you have deployed a SSO server instance beforehand", but I can see how people can misinterpret that and think that it binds to "you can skip the SSO server deployment".

Something like this might be better:

> If you deployed an SSO server beforehand, e.g. by executing `oc create -f service.sso.yaml`, set the system property `skip.sso.init` to `true` when running the test.
> The test will then use the existing SSO server and won't deploy its own.
> mvn clean verify -Popenshift,openshift-it -Dskip.sso.init=true

Ladicek:
Otherwise LGTM now.

rhoads-zach:
[test]

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
```diff
@@ -0,0 +1,30 @@
+[#running-the-secured-booster-integration-tests_context]
```
Ladicek: Ah! Should `context` here be actually `{context}`?

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+. In a terminal application, navigate to the directory with your project.
+. Execute the integration tests:
++
+[source,bash,options="nowrap",subs="attributes+"]
+--
+mvn clean verify -Popenshift,openshift-it
+--
+
+If you deployed an SSO server beforehand, e.g. by executing `oc create -f service.sso.yaml`, set the system property `skip.sso.init` to `true` when running the tests (example shown below).
```
Ladicek: Parens are a big no-no in our docs :-)

We can do something like this, though:

```
If you deployed an SSO server beforehand, e.g. by executing `oc create -f service.sso.yaml`, set the system property `skip.sso.init` to `true` when running the tests:

--
mvn clean verify -Popenshift,openshift-it -Dskip.sso.init=true
--

When executed like this, the tests will use the existing SSO server.
The tests will not deploy their own SSO server, nor will they destroy the existing one.
```

WDYT?

spisiakm: Sounds good, I'll update it right away if no one has any objections :)

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
spisiakm: I don't know (nothing really changed when I built it with `{context}`). If so, then also the `proc_running-the-secured-booster-integration-tests-on-sb.adoc` needs to be updated, as it wasn't there before.

Ladicek: It was probably wrong, but I'm not 100% sure. We'll need someone from the docs team to confirm.

Ladicek:
Can I trigger tests too?

[test]

Ladicek:
Seems I can't. @rhoads-zach could you please take a look? Thanks.

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
inoxx03: @spisiakm I think @Ladicek has a good point, `context` here should be `{context}`, that is, it should be marked up as a variable, as this is a measure that ensures module  reusability while avoiding section ID conflicts. The reason it does not get rendered every time is that we define it as an empty variable if  the module is not being used multiple times. If it is not defined in the `master.adoc` file it will simply not show up. but it's still there:) Does that make sense?

Ladicek: @inoxx03 Thing is, the files are now runtime-specific and there's no reuse anymore. Do you still use `{context}` in such situations? Thanks!

rhoads-zach:
[test]

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
inoxx03: If the module is runtime specific and only used i one place, then I'd be inclined to omit `{context}` and use the designator for the appropriate runtime. in this case, the ID would be: `[#running-the-secured-booster-integration-tests_vertx]`

rhoads-zach:
@Ladicek tests look like they now pass. I'll see if i can get you added to invoke tests

docs/topics/proc_running-the-secured-booster-integration-tests-on-vx.adoc
spisiakm: @rhoads-zach @Ladicek If I change `context` to `{context}`, the script `validate_guides.sh` fails to validate vertx-runtime:
```shell
master.xml:3499: element section: Schemas validity error : 
Element '{http://docbook.org/ns/docbook}section', 
attribute '{http://www.w3.org/XML/1998/namespace}id': 'running-the-secured-booster-integration-tests_{context}' is not a valid value of the atomic type 'xs:ID'.
```
The same applies to spring-boot-runtime (wrapping `context` in braces).

Ladicek: I think, given that those files are no longer shared, we can just hardcode the value. So instead of `context`, use `sb` or `vertx` or something like that.

spisiakm: +1, I would use just `sb` or `vertx` as well. I'll push an update if nobody is against it.

inoxx03: +1 to using hardcoded values if the module is not reused. @spisiakm FYI The reason why the validation fails when `{context}` is wrapped in curly braces is likely because the ID can not contain a variable that is not defined. It can however contain an empty variable. So, as a workaround, you can set `:context:` to be empty in the master file. this should fix the issue. That being said, please use hardcoded values i this specific instance :)

spisiakm:
@inoxx03 thanks for an explanation :)

rhoads-zach:
[test]

PR #1119, opened 2018-05-11
minor cleanup

rhoads-zach:
@inoxx03 @tradej let me know what you think.


docs/landing-page/master.adoc
```diff
@@ -69,7 +69,7 @@ include::topics/templates/document-attributes.adoc[]
     <img src="/docs/images/launch_logo.svg" alt="Fabric8 Launcher Logo" class="pull-right">
     <h2>Fabric8 Launcher documentation</h2>
     <p class="lead">
-      Read how to get started and setup your environment and all the good stuff that you can do with Fabric8 Launcher.
+      Read how to set up your environment, use the Fabric8 Launcher, and work with cloud-native runtimes on OpenShift.
```
tradej: We should avoid phrasal verbs, so I suggest replacing _set up_ with _configure._

docs/nodejs-runtime/master.adoc
```diff
@@ -18,7 +18,7 @@ include::topics/dev-guide-mission-intro.adoc[leveloffset=+1]
 [[nodejs-available-missions-and-boosters]]
 == Available Missions and Boosters for {NodeJS}
 
-The {NodeJS} runtime supports the following missions and boosters.
+The {NodeJS} runtime has the following missions and boosters.
```
tradej: The word _Has_ implies ownership, and the runtime can't really own things. I suggest using

_The following missions and boosters are available for the {NodeJS} missions:_

On a related note, why do we separate missions and boosters in this sentence? It's not factually correct, as it's actually just the boosters that we are talking about, or is it?

docs/landing-page/master.adoc
Ladicek: I don't know, you're the wordsmiths here :-)

docs/nodejs-runtime/master.adoc
inoxx03: @tradej @Ladicek  


> The following missions and boosters are available for the {NodeJS} missions

sounds a bit redundant (also, it makes no sense to say that we have missions available for missions), I'd go with something like:

> The following boosters are available for the {NodeJS} missions

alternatively:

> The following boosters are available for {NodeJS}

IMHO, there is little point in mentioning missions that do not have boosters available in a particular runtime. Therefore, saying there are boosters available for Node implies the missions that said boosters are base on are available, too.
Does that make sense or did I just spectacularly miss the point:?)

docs/landing-page/master.adoc
inoxx03: @tradej From a  stylistic POV, I agree, phrasal verbs should be used with caution. But, AaMoF, I think _set up_ is fairly acceptable here, too. My experience is that we typically refer to _configuring_ applications, for instance. But as @Ladicek says, this wording implies a broader set of actions. That being said, I'm all for stylistic consistency, and since neither of the options here are wrong, I'd go with the more stylistically acceptable one, which would be _configure_.

docs/nodejs-runtime/master.adoc
Ladicek: The entire "launching space ships" metaphor just doesn't work all that well...

I agree with _The following boosters are available for {NodeJS}_, it's the simplest and also factually most correct.

docs/landing-page/master.adoc
```diff
@@ -98,7 +98,7 @@ include::topics/templates/document-attributes.adoc[]
       <div class="col-sm-3 runtime">
         <div class="fa icon vertx"></div>
         <p >
-          Learn how to use launcher for Eclipse Vert.x projects
+          Learn how to use the Eclipse Vert.x runtime.
```
Ladicek: When I suggested to simplify "Learn how to use the WildFly Swarm runtime" to "Learn how to use WildFly Swarm", I thought it would be obvious that the suggestion also applies to all the other runtimes :-)

Ladicek:
Otherwise LGTM.

docs/landing-page/master.adoc
Ladicek: With the build app from scratch, remote debugging or debug logging chapters, I feel all the runtime guides aspire to be general-purpose documentation. I can see what you're saying, but I don't think it will be obvious to a regular reader.

rhoads-zach: @Ladicek fair point, updating.

rhoads-zach:
@Ladicek amended

PR #1126, opened 2018-05-17
fixes RHOARDOC-1309

inoxx03:
Added procedure to change property names to ensure Agroal works.


inoxx03:
@rhoads-zach @rdruss @spisiakm Please let me know what you think

rdruss:
Given that there was a note in the previous section about setting the provider_class when using c3p0, it might make sense to also include this in the example for c3p0

rhoads-zach:
@rdruss thank you.

@inoxx03 merging this for now. Please provide a follow up PR to address the above feedback.

PR #1127, opened 2018-05-18
follow-up to RHAORDOC-1298

inoxx03:
* included property to explicitly set C3P0 as the data source provider in the C3P0 config example
* added comment to the C3P0 example to show where the extra property is set
* minor wording fixes: (c3p0 ->  C3P0) + a few others
* hardcoded _RHOAR_ as the product short name since `{ProductShortName}` maps to upstream name


Ladicek:
Neither C3P0 nor Agroal are "JDBC connection client"; correct term is "JDBC connection pool" (or just "connection pool").

inoxx03:
@lthon thank you, updating

inoxx03:
@rhoads-zach please feel free to merge if this looks OK to you.

PR #1128, opened 2018-05-21
changes base url. fixes RHOARDOC-1118

rhoads-zach:


rhoads-zach:
@Ladicek 

All of these files one of the following, and require a hard-coded reference:
* readme file
* property file used by readme
* file used directly by the front end (for now)

Ladicek:
I know that some of the files are READMEs or properties, but I thought some of them are not. Sorry if I was wrong :-)

tradej:
I understand the README file and the property files, but I thought the Launcher did include `document-attributes.adoc`. Anyway, if that's not the case, the change looks good.

Two things:

* I was not able to check the liveness because the files are not up yet.
* I am getting an SSL bad certificate domain error when I access any of the guide URLs. I presume it will go away when the files are up, but I wouldn't like for it to fall through the cracks if it doesn't.

rhoads-zach:
amended PR with latest changes.

New base url: https://launcher.fabric8.io

@gastaldi @tradej @inoxx03 @Ladicek let me know what you think

tradej:
LGTM.

PR #1130, opened 2018-05-23
RHOARDOC-1317: Fixed WFS attribute

tradej:
Comments do not evaluate, so the whole line was set as the attribute


tradej:
Two of us, @Ladicek, two of us.

PR #1132, opened 2018-06-05
RHOARDOC-867: Consolidated module names, anchor IDs and file names

tradej:
The change does the following:

* For all the procedure modules (files that more-or-less comply with the Modular Documentation Guidelines), it takes the heading and modifies the anchor ID and file name to match it.
  * The notable exception to that rule is the usage of certain attributes in headings like `name-mission-*` or `parameter-*-name`, which are not used in the anchor IDs and file names. The ugliness isn't worth it IMO. If you think we should stick to the rules, though, I will fix it and regenerate the docs.
  * All references (`xref`s) and include directives are updated to match too.
* Adds the `_{context}` suffix to all anchor IDs in the modules. The suffix will likely start to be mandated in the Modular guidelines, so it's a step we may want to do anyway.
  * All the references are updated to match too.
* Certain parameters needed to be updated to match the new file names.
* Some headings were modified to avoid clashes with existing or identically-named modules.

This PR partially resolves RHOARDOC-867.


Ladicek:
I'm gonna abstain :-) Will verify as part of the diff check.

Ladicek:
You mean `rest-http`, not `rest-api`? (To be frank, `http-api` is by far the best description of what it is, but that ship has long sailed.)

tradej:
@rhoads-zach What occurrence are you referring to? If it's the `{parameter-mission}` value, that should not be visible to the user in the rendered document AFAICT as it is only used in an `include` directive in the `assembly_building-and-deploying-the-booster-to-openshiftlocal` assembly to select a file name.

rhoads-zach:
@Ladicek you're correct.

@tradej comment withdrawn, i double checked something and the link is unaffected

rhoads-zach:
@tradej is this good to merge?

tradej:
@rhoads-zach To my best knowledge, yes.

BTW this will be followed upon with other PRs much to the same effect (with concepts, references, assemblies, and other files currently not modular).

rhoads-zach:
thanks @tradej merging.

PR #1133, opened 2018-06-05
Added appendix on persisting deployment config. Fixes RHOARDOC-1321

rhoads-zach:
@tradej @inoxx03 Please take and let me know what you think. I thought it made the most sense to put this as an appendix, but i'm open to reworking/including somewhere else if you have a better idea.


docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
+[[proc_updating-booster-deployment-configuration]]
```
tradej: The heading should be `[id='updating-the-deployment-configuration-of-a-booster_{context}']` (Let's use the `_{context}` suffix everywhere since it's likely going to be mandated by the guidelines.)

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+= Updating the Deployment Configuration of a Booster
+
+The deployment configuration for a booster contains information related to deploying and running the booster in OpenShift, such as route information or readiness probe location. 
+The deployment configuration of a booster is stored in a set YAML files.
```
tradej: A set _of_ YAML files?

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
+[[proc_updating-booster-deployment-configuration]]
+= Updating the Deployment Configuration of a Booster
+
+The deployment configuration for a booster contains information related to deploying and running the booster in OpenShift, such as route information or readiness probe location. 
+The deployment configuration of a booster is stored in a set YAML files. 
+For boosters that use the Fabric8 Maven Plugin, the YAML files are located in the `src/main/fabric8/` folder. For {NodeJS} boosters using Nodeshift, the YAML files are located in the `.nodeshift` folder.
```
tradej: Use _directory_ instead of _folder._ (That's Windows terminology.)

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+.Prerequisite
+* An existing booster project.
+* The `oc` CLI client installed.
+
+.Procedure
+
+. Edit the an existing YAML file or create an additional YAML file with your configuration update.
```
tradej: Remove _the._

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+** For example, if your booster already has a YAML file with a `readinessProbe` configured, you could change `path` value to different available path to check for readiness:
```
tradej: ..._the_ `path` value...

...to _a_ different available...

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+    spec:
+      containers:
+        readinessProbe:
+          httpGet:
+            path: /path/to/probe
+            port: 8080
+            scheme: HTTP
+...
+----
+** If a `readinessProbe` is not configured in an existing YAML file, you can also create a new YAML in the same directory with the `readinessProbe` configuration.
```
tradej: ...create a new YAML _file_...

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+. Deploy the updated version of your booster using Maven or `npm`.
```
tradej: I wouldn't put npm in verbatim because Maven isn't. Not that it's wrong, it just looks better IMO.

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,72 @@
...
+            initialDelaySeconds: 60
+            periodSeconds: 30
+            successThreshold: 1
+            timeoutSeconds: 1
+          ...
+----
+
+
+.Related Information
+If you updated the configuration of your application directly using the web-based console or the `oc` CLI client, you must export and add these changes to your YAML file.
```
tradej: I'd argue for dropping _you must_ because it's doesn't really add value IMO. (minimalism)

docs/topics/proc_updating-booster-deployment-configuration.adoc
Ladicek: I don't know about Nodeshift, but there's a very important point when it comes to Fabric8 Maven plugin: the YAML files stored in `src/main/fabric8` can, but _don't have to_ be the full OpenShift resource definitions. In case of the boosters, they often _aren't_! FMP is able to fill in the missing parts. Which can be convenient, but also confusing and even misleading (for example, if you specify a `Deployment` but don't add the _full_ specification, FMP actually changes that to a `DeploymentConfig` when it knows it's targetting OpenShift!).

Not sure how to express that in docs, but it's pretty important.

inoxx03: I think Mac uses _folder_, too.

inoxx03: @Ladicek Would you know if this opinionated feature of FMP is mentioned anywhere in the FMP docs? I'll take a look and if I find any such information documented there, I'd propose we link to that explanation. On the other hand, if there is no mention of it anywhere, I'd be happy to write something up to add that information, since it is IMHO important.

rhoads-zach:
@tradej @inoxx03 @Ladicek updated
FYI @lholmquist added callout about how FMP and nodeshift uses deployment configs. let know if you have any feedback.

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,74 @@
+[id='proc_updating-booster-deployment-configuration_{context}']
+= Updating the Deployment Configuration of a Booster
+
+The deployment configuration for a booster contains information related to deploying and running the booster in OpenShift, such as route information or readiness probe location. 
+The deployment configuration of a booster is stored in a set of YAML files. 
+For boosters that use the Fabric8 Maven Plugin, the YAML files are located in the `src/main/fabric8/` directory. For boosters using Nodeshift, the YAML files are located in the `.nodeshift` directory.
+
+IMPORTANT: The deployment configuration files used by the Fabric8 Maven Plugin and Nodeshift do not have to be full OpenShift resource definitions. Both Fabric8 Maven Plugin and Nodeshift can take the provided files and infer some missing information to create a full OpenShift resource definition when targeting OpenShift.
```
Ladicek: Is it OK in case of a call-out to put multiple sentences on a single line? :-)

No, this looks good to me. I think we should probably suggest to inspect the full definitions as generated by FMP/Nodeshift, or at least make a note where they can be found. In case of FMP, the generated resources are in `target/classes/META-INF/fabric8`.

Ladicek: Actually one more: it's not 100% true that they can infer missing info just _when targeting OpenShift_. Again not sure about Nodeshift, but in case of FMP, it always generates both OpenShift and Kubernetes resources, and it adds missing info _always_ (even when targetting plain Kubernetes). I'd just remove the words "when targetting OpenShift".

lholmquist: nodeshift just targets openshift

Ladicek: Good point and I think removing the last 3 words ("when targetting OpenShift") won't change the meaning for Nodeshift.

rhoads-zach:
@Ladicek updated with locations for FMP and nodeshift generated resources.

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,81 @@
+[id='proc_updating-booster-deployment-configuration_{context}']
```
tradej: As per my previous comment, the ID should be `[id='updating-the-deployment-configuration-of-a-booster_{context}']`.

tradej: Oh, and I forgot that the file name should be `proc_updating-the-deployment-configuration-of-a-booster.adoc`.

docs/topics/proc_updating-booster-deployment-configuration.adoc
```diff
@@ -0,0 +1,81 @@
...
+[IMPORTANT]
+====
+The deployment configuration files used by the Fabric8 Maven Plugin and Nodeshift do not have to be full OpenShift resource definitions. 
+Both Fabric8 Maven Plugin and Nodeshift can take the provided files and infer some missing information to create a full OpenShift resource definition.
+The resource definitions generated by the Fabric8 Maven Plugin are available in the `target/classes/META-INF/fabric8/` directory.
+The resource definitions generated by Nodeshift are available in the `tmp/nodeshift/resource/` directory.
+====
+
+
+.Prerequisite
```
tradej: Sorry, missed that before. Should be _Prerequisites_ as there are more than one.

rhoads-zach:
@tradej updated, thank you for the feedback

PR #1135, opened 2018-06-11
RHOARDOC-867: Consolidated naming of concepts and concept-like files

tradej:
Similar content to #1132, only with files that are now concept modules.

I have moved certain mission intro files' contents to the `master.adoc` file. That is a mid-step before I move all content related to respective missions into assemblies, and this content will effectively become assembly introduction.

Some files (notably concerning the REST API and Secured mission) are untouched because with those, a more thorough refactoring is necessary due to the layout of the content.

PR #1136, opened 2018-06-11
fixed build error with {context} not being set

rhoads-zach:
@tradej mind taking a quick look?


tradej:
Instead of setting the context for a single file, I have removed the unset command in dd34baec3b93e5a69867224c6b12016ea05f5d02, so the build now succeeds.

Thank you for the PR, but I believe it's not necessary any more since #1135 was merged.

rhoads-zach:
thank you @tradej closing

PR #1137, opened 2018-06-13
RHOARDOC-867: Consolidated naming of references and reference-like files

tradej:
Similar content to #1132, only with files that are now reference modules.

I have not used the `_{context}` suffix in the appendices because they are referenced from various contexts, and there is virtually zero chance of appendix reuse inside one guide.


Ladicek:
Once again, I'm gonna abstain and will verify as part of the diff check :-)

tradej:
No problem, thank you. :)

PR #1139, opened 2018-06-13
added script to search for unused files

rhoads-zach:
@tradej @inoxx03 @gastaldi let me know what you think.


scripts/list_unused_topics.sh
```diff
@@ -0,0 +1,26 @@
...
+cd $DOCS_SRC
+master_files_names=`find */master.adoc`
+
+cd $TOPICS_SRC
+topic_file_names=`ls *.adoc`
+
+cd $DOCS_SRC
+for name in $topic_file_names
+do
+  if [[ $(grep $name $SKIP) ]]; then
```
tradej: This could be, I believe, replaced with `if grep -q $name $SKIP; then`.

scripts/list_unused_topics.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+CURRENT_DIR="$( pwd -P)"
+SCRIPT_SRC="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
+TOPICS_SRC="$( dirname $SCRIPT_SRC )/docs/topics"
+DOCS_SRC="$( dirname $SCRIPT_SRC )/docs"
+SKIP="$SCRIPT_SRC/skip_unused_files.txt"
+
+cd $DOCS_SRC
```
tradej: I recommend using `pushd` instead of `cd` everywhere because if you interrupt the script halfway through, you'll automatically be returned to the previous directory in the directory stack, i. e. your original working directory. (I may be wrong here, but I think it works this way)

scripts/list_unused_topics.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+CURRENT_DIR="$( pwd -P)"
```
tradej: Seems unused.

scripts/list_unused_topics.sh
```diff
@@ -0,0 +1,26 @@
...
+SCRIPT_SRC="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
+TOPICS_SRC="$( dirname $SCRIPT_SRC )/docs/topics"
+DOCS_SRC="$( dirname $SCRIPT_SRC )/docs"
+SKIP="$SCRIPT_SRC/skip_unused_files.txt"
+
+cd $DOCS_SRC
+master_files_names=`find */master.adoc`
```
tradej: Why not `ls */master.adoc`, or rather `find -name master.adoc`?

scripts/list_unused_topics.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cd $TOPICS_SRC
+topic_file_names=`ls *.adoc`
+
+cd $DOCS_SRC
+for name in $topic_file_names
```
tradej: I think this may fail with file names that contain spaces.

scripts/list_unused_topics.sh
rhoads-zach: did some double checking and this works. fixed.

scripts/list_unused_topics.sh
rhoads-zach: @tradej FYI, `pushd` causes 3 lines of output at the beginning that are relative paths to the search dirs:

```bash
~/github/documentation/docs ~/github/documentation
~/github/documentation/docs/topics ~/github/documentation/docs ~/github/documentation
~/github/documentation/docs ~/github/documentation/docs/topics ~/github/documentation/docs ~/github/documentation
```

I'll see if i can work around this. 

Otherwise, output is identical.

rhoads-zach: found a workaround.

scripts/list_unused_topics.sh
rhoads-zach: removed

scripts/list_unused_topics.sh
rhoads-zach: fixed

rhoads-zach:
@tradej let me know what you think

tradej:
LGTM.

PR #1140, opened 2018-06-13
replaces aaaaa with UID. Fixes RHOARDOC-17

rhoads-zach:


Ladicek:
I honestly think this is a change for the worse.

PR #1142, opened 2018-06-20
Clarified prequisite for deploying a booster to OCP. Fixes RHOARDOC-1347

rhoads-zach:
@Ladicek let me know what you you think.

### DIFF
--- a/docs/topics/proc_deploying-the-booster-to-openshiftcontainerplatform.adoc
+++ b/docs/topics/proc_deploying-the-booster-to-openshiftcontainerplatform.adoc
@@ -11,7 +11,7 @@ The process of creating and deploying boosters to {OpenshiftContainerPlatform} i
 
 .Prerequisites
 
-* The booster created on either xref:deploying-the-booster-to-openshiftlocal_{context}[{OpenShiftLocal}] or xref:deploying-the-booster-to-openshiftonline_{context}[{OpenShiftOnline}].
+* The booster created using link:{link-launcher-oso}[{launcher-oso}] or link:{link-openshift-local-guide}[the {launcher} tool].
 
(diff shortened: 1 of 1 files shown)


docs/topics/proc_deploying-the-booster-to-openshiftcontainerplatform.adoc
```diff
@@ -11,7 +11,7 @@ The process of creating and deploying boosters to {OpenshiftContainerPlatform} i
 
 .Prerequisites
 
-* The booster created on either xref:deploying-the-booster-to-openshiftlocal_{context}[{OpenShiftLocal}] or xref:deploying-the-booster-to-openshiftonline_{context}[{OpenShiftOnline}].
+* The booster created using link:{link-launcher-oso}[{launcher-oso}] or link:{link-openshift-local-guide}[the {launcher} tool].
```
Ladicek: You mean a .zip here, right? Because you can't deploy to OCP from Launcher, unless you deploy Launcher to OCP first.

rhoads-zach: @Ladicek I wanted to narrow the language to focus on 'creating with the launcher'. From there a user can then take that artifact and deploy it to OCP. I realize a side effect of doing the github flow on either OSO or minishift will result in the booster being deployed to that same instance, but the user can still clone the repo and deploy it to OCP. They are not forced to use the zip method.

Does this make sense or do you think the point is too subtle?

Ladicek: Ah true. Right, sounds OK.

PR #1144, opened 2018-06-20
Fixes sboot version. Fixes RHOARDOC-1349

rhoads-zach:
@inoxx03 @Ladicek let me know what you think.


rhoads-zach:
@inoxx03 @Ladicek please ignore, was named incorrectly. fixed in #1145

PR #1148, opened 2018-06-20
reworked awkward titles. Fixes RHOARDOC-1352

rhoads-zach:
@Ladicek let me know what you think


rhoads-zach:
@Ladicek fixed, let me know what you think

PR #1159, opened 2018-07-19
added note about using the redhat base image

rhoads-zach:
@Ladicek @tradej @inoxx03 let me know what you think


Ladicek:
Yay! LGTM post-facto :-)

PR #1169, opened 2018-07-25
Added a description of the "half open" state.

huntley:
@inoxx03  @rhoads-zach 
This change needs a decent review. While I believe the update is correct, there may be other sections that are affected.

### DIFF
--- a/docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
+++ b/docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
@@ -1,5 +1,6 @@
 // This is a parameterized module. Parameters used:
 //
+//
 //  parameter-mission: id of the mission. This is used in anchors file imports.
 //  parameter-mission-name: used in the mission title
 //
@@ -45,6 +46,7 @@ The state can be:
 ifdef::vert-x[]
 ** _half-open_: the circuit breaker is allowing a request to reach the service. If the request succeeds, the state of the service is reset to closed. If the request fails, the timer is restarted.
 endif::[]
+
 The following steps demonstrate how to verify the availability of the service, simulate a failure and receive a fallback response.
 
 //TODO: add a warning not to use `http` as it may contain cached responses from the remote endpoint.
--- a/docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
+++ b/docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
@@ -42,7 +42,9 @@ If the remote endpoint is unavailable, the `name` service responds with an HTTP 
 The state can be:
 ** _open_ : the circuit breaker is preventing requests from reaching the failed service,
 ** _closed_: the circuit breaker is allowing requests to reach the service.
-
+ifdef::vert-x[]
+** _half-open_: the circuit breaker is allowing a request to reach the service. If the request succeeds, the state of the service is reset to closed. If the request fails, the timer is restarted.
+endif::[]
 The following steps demonstrate how to verify the availability of the service, simulate a failure and receive a fallback response.
 
 //TODO: add a warning not to use `http` as it may contain cached responses from the remote endpoint.


tradej:
ok to test

tradej:
add to whitelist

docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
```diff
@@ -41,7 +41,8 @@ If the remote endpoint is unavailable, the `name` service responds with an HTTP
 * the `/api/cb-state` endpoint, which returns the state of the Circuit Breaker.
 The state can be:
 ** _open_ : the circuit breaker is preventing requests from reaching the failed service,
-** _closed_: the circuit breaker is allowing requests to reach the service.
+** _closed_: the circuit breaker is allowing requests to reach the service,
+** _half open_: the circuit breaker is allowing a request to reach the service. If the request succeeds, the state of the service is reset to open. If the request fails, the timer is restarted.
```
Ladicek: While circuit breakers can be in the half-open state, the `/api/cb-state` endpoint actually only ever returns `open` or `closed`.

inoxx03: @huntley Nicely written, but I have a few comments:
1. The _open_ state of the CB is the one that prevents traffic to the protected service. Therefore, you should change

> If the request succeeds, the state of the service is reset to open

to something like:

> If the request succeeds, the state of the circuit breaker is reset to _closed_

2. The half-open CB typically allows multiple initial requests to reach the recovering service, rather than just one. I'd say the point is to allow limited traffic to the service, while preventing it from being flooded with requests immediately upon recovery.
3. If any of the limited amount of initial requests fails, the circuit breaker goes into an open state (which also implies that the recovery timeout counter is restarted)

Ladicek: You're very much right here, but please see my comment :-)

huntley: Good point. Thank you!

inoxx03: @huntley @Ladicek after following up with @akoniar of Vertx QE, it seems that the Vertx CB booster has a `half-open` state.

I have found some references to it in the [integration tests|https://github.com/openshiftio-vertx-boosters/vertx-circuit-breaker-booster/blob/master/integration-tests/src/test/java/io/openshift/booster/OpenShiftIT.java#L101] and the [HTML frontend|https://github.com/openshiftio-vertx-boosters/vertx-circuit-breaker-booster/blob/master/greeting-service/src/main/resources/webroot/index.html#L139] of the booster.

inoxx03: @Ladicek  @akoniar confirms that the `/api/cb-state/` endpoint of the Vertx booster does indeed return a `half-open` state. Thisi specific to the Vertx booster.
@huntley Seems that we need to wrap the section describing the `half-open` state  in a conditional statement, so it only appears in the Vertx guide. Please let me know if you'd like some more info about how you can do it.

huntley:
Fixed open issues and verified that the vert.x content does not appear in the other runtime docs. Hopefully all set.

docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
```diff
@@ -42,7 +42,9 @@ If the remote endpoint is unavailable, the `name` service responds with an HTTP
 The state can be:
 ** _open_ : the circuit breaker is preventing requests from reaching the failed service,
 ** _closed_: the circuit breaker is allowing requests to reach the service.
-
+ifdef::vert-x[]
+** _half-open_: the circuit breaker is allowing a request to reach the service. If the request succeeds, the state of the service is reset to closed. If the request fails, the timer is restarted.
+endif::[]
```
tradej: I believe there should be an empty line below this statement to make sure that the sentence on line 48 is not appended to the _half-open_ bullet point in the Vert.x guide.

huntley: Dang nab it. Good catch.

rhoads-zach:
LGTM, thank you @huntley

docs/topics/proc_interacting-with-the-circuit-breaker-booster.adoc
huntley: All set now (fingers crossed)

PR #1170, opened 2018-08-10
fixes RHAORDOC-1390

inoxx03:
updated references to latest Vertx version and Vertx Maven plugin

@huntley FYI^^

### DIFF
--- a/docs/topics/templates/document-attributes.adoc
+++ b/docs/topics/templates/document-attributes.adoc
@@ -104,8 +104,8 @@
 :SpringBootMvnPluginVersion: 1.5.14.RELEASE
 
 // used in the BOM file example and in the Vert.x Runtime developer Guide
-:VertXVersion: 3.5.1.redhat-004
-:VertXMvnPluginVersion: 1.0.15
+:VertXVersion: 3.5.3.redhat-00001
+:VertXMvnPluginVersion: 1.0.17
 
 :nodeshiftNodeVersion: --dockerImage=registry.access.redhat.com/rhoar-nodejs/nodejs-8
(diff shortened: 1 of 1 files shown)


docs/topics/templates/document-attributes.adoc
```diff
@@ -104,8 +104,8 @@
 :SpringBootMvnPluginVersion: 1.5.14.RELEASE
 
 // used in the BOM file example and in the Vert.x Runtime developer Guide
-:VertXVersion: 3.5.1.redhat-004
-:VertXMvnPluginVersion: 1.0.15
+:VertXVersion: 3.5.3.redhat-00001
```
rhoads-zach: @inoxx03 is `00001` a typo?

huntley: Do not believe so, that is the version in the BOM though I know it is different from previous releases.

inoxx03: @rhoads-zach Can confirm, not a typo. I've checked the version against the BOM file and it checks out. Not sure why the string changed, could be due to tooling updates on the Vertx productization side.

rhoads-zach: thank you @huntley and @inoxx03

rhoads-zach:
@inoxx03 @huntley can i merge this?

docs/topics/templates/document-attributes.adoc
Ladicek: This is actually a new productization convention. All products will have the numeric suffix in a 5-digit form.

PR #1175, opened 2018-08-15
RHOARDOC-1393: WildFly Swarm rename to Thorntail

tradej:
This is a set of tightly related changes that are tracked in the sub-tasks of [RHOARDOC-1393](https://issues.jboss.org/browse/RHOARDOC-1393).

**Notes:**

* The `docs/topics/wildfly-swarm` directory was abandoned in favour of `docs/topics/thorntail`.
* The new version of Thorntail was synced because the naming changes were not backported.
* The references to WFS remain in the Contribution Guide topics because they're in the same commit as the sync script changes.
* The guide directory was renamed to `docs/thorntail-runtime`.
* The guide file was changed to `thorntail-runtime.html` and I have set up redirects from the old guide to the new one. I have tested them on the local server, but please give them go as well.
* I have added an admonition at the beginning of the guide with the info about the rename.
* I have added a workaround for ENTSWM-458 to the sync script, which prevents upstream syncing. It does not affect our docs delivery in any way at the moment.

**Changes currently missing from this PR:**

* The changes to the sync script (see [RHOARDOC-1394](https://issues.jboss.org/browse/RHOARDOC-1394)) are missing from this PR for productisation reasons. The commit with those will be provided later in a separate PR.
* The new Thorntail logo is missing from the launcher landing page. I will add it when it's published.
* The URLs to the community documentation will be provided when it's published.

The changes are pretty strictly organised into commits, so in my opinion, it makes sense to review individual commits as opposed to the entire change set at once.

Resolves RHOARDOC-1393, 1398, 1399, 1400, and 1358.


tradej:
Will do, @gastaldi, thanks.

huntley:
LGTM

tradej:
Thank you, @huntley.

nginx.conf
```diff
@@ -57,15 +57,25 @@ http {
         }
 
         # Redirects from the old URLs generated by the boosters
-        location ~* /docs/mission-([^/]+)-(wf-swarm|vertx).html {
-            return 301 /docs/$2-runtime.html#mission-$1-$2;
+        location ~* /docs/mission-([^/]+)-vertx.html {
+            return 301 /docs/vertx-runtime.html#mission-$1-$2;
```
Ladicek: There's no `$2` anymore here.

tradej: Great catch, will fix. Thanks.

tradej:
@Ladicek The preview is [here](http://download.eng.brq.redhat.com/scratch/tradej/rhoar/thorntail-rename/docs/thorntail-runtime.html) (VPN required). I'm especially not sure of some Artifact IDs that still say `swarm`, so please advise.

Ladicek:
Most of the occurences of `swarm` are OK, because all the Java package names are still `org.wildfly.swarm.*` and all the configuration keys are still `swarm.*`. But there are indeed some outdated pieces:

- In chapter _4.1.1. Creating an application from scratch_, in the listing of `pom.xml`, we have:
    - `<name>Wildfly Swarm Example</name>` -- that should read `<name>Thorntail Example</name>`;
    - `version.wildfly.swarm` -- that should be `version.thorntail` (the value of that property will change to something like `2.2.0.Final-redhat-00015`, but we don't know the exact version number yet);
    - a XML comment `<!-- Wildfly Swarm Fractions -->` -- I'd actually remove it completely, but if not, then it needs to change to `<!-- Thorntail Fractions -->`.
- In chapter _4.2. Deploying an existing Thorntail application to OpenShift_, in the listing of `pom.xml`, we have:
    - `<include>wildfly-swarm</include>` -- that should read `<include>thorntail-v2</include>`;
    - `<wildfly-swarm-health-check>` -- that should read `<thorntail-v2-health-check>`.
- In chapter _D.23.5. MicroProfile Metrics_, we have a hardcoded reference to `org.wildfly.swarm:microprofile-metrics` at the beginning. This needs to change to `io.thorntail:microprofile-metrics`. Also a few paragraphs below, the example says `swarm-demo` -- that should probably be changed to `thorntail-demo`.
- The warning at the very beginning of chapter _D.24. Monitor_ includes `org.wildfly.swarm:microprofile-health`. That needs to change to `io.thorntail:microprofile-health`.

That's all I could find.

tradej:
Thank you for the feedback, @Ladicek. Remaining problems fixed in https://github.com/thorntail/thorntail/pull/1139 and https://github.com/wildfly-swarm-prod/wildfly-swarm/pull/150; please merge them, I will resync this PR when those are accepted.

Ladicek:
Cool, thanks, will merge the PRs once the PR jobs finish.

PR #1176, opened 2018-08-16
Fixed RHOARDOC-1407: fixed bug causing CRUD nodejs interaction to not…

rhoads-zach:
… display


rhoads-zach:
@Ladicek @tradej @inoxx03 let me know what you think

PR #1180, opened 2018-08-21
RHOARDOC-1406: updated minishift addon section title

rhoads-zach:


rhoads-zach:
@Ladicek updated title per feedback. Let me know what you think.

PR #1184, opened 2018-08-29
Fixed rendering issue of callout

huntley:
Should be a minor change but need to check PDF output to verify that the fix works in both HTML and PDF.


inoxx03:
Add to whitelist

PR #1185, opened 2018-08-30
Fixes for the RHOARDOC-1415 pull request

tradej:


docs/topics/con_how-caching-works-and-when-you-need-it.adoc
```diff
@@ -1,7 +1,7 @@
 [id='how-caching-works-and-when-you-need-it_{context}']
 = How caching works and when you need it
 
-Caching allows you to store information and access it for a given period of time.
+Caches allows you to store information and access it for a given period of time.
 Caching can provide access to information faster or more reliably than repeatedly calling the original service.
```
Ladicek: This line has the same problem :-)

PR #1197, opened 2018-10-15
RHOARDOC-1455: Fixed tag in POM

tradej:
Note: The added files are inconsequential as they are not used in the Thorntail Runtime Guide, only the modified POM files at the bottom of the diff are important.

Resolves RHOARDOC-1455.


tradej:
Thanks, merging.

PR #1202, opened 2018-10-25
Fixes RHAORDOC-1443: add Prometheus procedure module for Node.js

inoxx03:


docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,128 @@
...
+----
+$ curl http://localhost:8888/api/greeting
+{"content":"Hello, World!"}
+----
+
+. Wait at least 15 seconds for the collection to happen, and view the metrics in the Prometheus web UI:
++
+--
+. Open the Prometheus UI at link:http://localhost:9090/[http://localhost:9090/^] and type `hello` into the _Expression_ box.
+. From the suggestions, select for example `application:hello_count` and click the _Execute_ button.
```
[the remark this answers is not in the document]
inoxx03: @Ladicek It seems that the metric is indeed exposed as `hello_count`, I don't see the `application:` prefix anywhere, though.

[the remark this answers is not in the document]
Ladicek: The `application:` prefix is also part of the metric name, so this needs fixing.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,129 @@
...
+. In the table below, you can see the number of times you invoked the resource method.
+// this metric seems to be unavailable for the node app.
+//. Alternatively, select `application:hello_time_mean_seconds` to see the mean time of all the invocations.
+
+//all user-defined metrics should have this
+Note that all metrics you created are prefixed with `application:`.
```
Ladicek: So the Node.js Prometheus library does that on its own? Cool.

inoxx03: @Ladicek Not sure it does, gonna have to check with Daniel to be certain.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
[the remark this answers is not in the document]
inoxx03: @Ladicek Seems like prom-client does not automatically ad metrics, but Dan has kindly updated the example app code to include all the application metrics exposed by Prometheus by default, as well as a custom application:prefix for the application metrics.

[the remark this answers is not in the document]
Ladicek: Yea, at this point I'd have to actually run it to provide meaningful feedback, and I'm too lazy to do that :-) But thanks!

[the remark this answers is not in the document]
inoxx03: @Ladicek Sure:) no worries:) I've run it to test the procedure and everything seems to check out:) Should be good:)

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,132 @@
...
+----
+
+Wait until the `Server is ready to receive web requests` message is displayed in the console.
+
+Include the the `--network host` option to allow Prometheus to connect to your application.
+--
+
+.Procedure
+
+. Install the `prom_client` npm package.
```
lance: Typo: this should be `prom-client`.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,132 @@
...
+const path = require('path');
+const express = require('express');
+const bodyParser = require('body-parser');
+const client = require('prom-client');
+
+// enable prom-client to expose default application metrics
+const collectDefaultMetrics = client.collectDefaultMetrics;
+
+// define a custom prefix string for application metrics
+collectDefaultMetrics({ prefix: 'application:' });
```
lance: I think the prefix should not have a colon.

An underscore and no space makes more sense to me and is congruent with the example from the `prom-client` module.

```js
client.collectDefaultMetrics({ prefix: 'application_'})
```

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+const histogram = new client.Histogram({
+  name: 'application:hello',
```
lance: It's really a nitpicky comment, I know, but here I would also use underscore. `application_hello`. As an aside, I might choose a different, more descriptive prefix :)

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
lance: @inoxx03 @Ladicek see line 77 of the node application. That line sets `application:` as the prefix.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,132 @@
...
+  help: 'Duration of HTTP requests in ms',
+  labelNames: ['status_code'],
+  buckets: [0.1, 5, 15, 50, 100, 500]
+});
+const app = express();
+const port = 8888;
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({extended: false}));
+app.use('/api/greeting', (request, response) => {
```
lance: Shouldn't this be `app.get('/api/greeting, ...)`?

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,132 @@
...
++
+--
+[source,bash,opts="nowrap"]
+----
+# docker run -it --rm --network host -p 9090:9090 -v $PWD/prometheus.yml:/etc/prometheus/prometheus.yml prom/prometheus
+----
+
+Wait until the `Server is ready to receive web requests` message is displayed in the console.
+
+Include the the `--network host` option to allow Prometheus to connect to your application.
```
lance: Note: this only works on Linux and is unsupported on Mac and Windows.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,134 @@
...
+----
+# docker run -it --rm --network host -p 9090:9090 -v $PWD/prometheus.yml:/etc/prometheus/prometheus.yml prom/prometheus
+----
+
+Wait until the `Server is ready to receive web requests` message is displayed in the console.
+
+When running your application in a Linux environment, include the the `--network host` option to allow Prometheus to connect to your application.
+
+NOTE: The `--network host` option is available only in Linux environments.
+It is not supported on Windows or MacOS.
```
Ladicek: There needs to be _some_ way for Prometheus to connect to the application. Using `--network host` is easiest, and actually I think the _only_ possible way if we keep the application running on `localhost`. Host networking is also used in the Thorntail variant of this guide.

If we care enough about Windows and Mac, we have to rework the entire guide to some extent.

Merely saying that host networking only works on Linux and that you should only add `--network host` when on Linux doesn't work.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,134 @@
...
+$ curl http://localhost:8888/api/greeting
+{"content":"Hello, World!"}
+----
+
+. Wait at least 15 seconds for the collection to happen, and view the metrics in the Prometheus web UI:
++
+--
+. Open the Prometheus UI at link:http://localhost:9090/[http://localhost:9090/^] and type `hello` into the _Expression_ box.
+. From the suggestions, select for example `application:hello_count` and click the _Execute_ button.
+Note that all metrics you created are prefixed with `application:`.
```
danbev: This is what I've been using locally on mac:
```yml
- job_name: 'nodejs'
    static_configs:
    - targets: ['192.168.1.80:8888']
```
```console
$ docker run -it --rm -p 9090:9090 -v $PWD/prom.yml:/etc/prometheus/prometheus.yml prom/prometheus
```
And then updating the app to listen on the ip address:
```javascript
'use strict';

const path = require('path');
const express = require('express');
const bodyParser = require('body-parser');
const client = require('prom-client');

const collectDefaultMetrics = client.collectDefaultMetrics;
collectDefaultMetrics({ prefix: 'application:' });

const histogram = new client.Histogram({
  name: 'application:hello',
  help: 'Duration of HTTP requests in ms',
  labelNames: ['status_code'],
  // buckets for response time from 0.1ms to 500ms
  buckets: [0.1, 5, 15, 50, 100, 500]
});

const app = express();
const port = 8888;

app.use(bodyParser.json());
app.use(bodyParser.urlencoded({extended: false}));

app.use('/api/greeting', (request, response) => {
  const end = histogram.startTimer();
  const name = request.query ? request.query.name : undefined;
  response.send({content: `Hello, ${name || 'World!'}`});
  end();
});

app.get('/metrics', (req, res) => {
  res.set('Content-Type', client.register.contentType);
  res.end(client.register.metrics());
})
const ip = '192.168.1.80';
app.listen(port, ip, () => console.log(`Example app listening on port ${ip}:${port}!`))
```

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
inoxx03: @Ladicek thank you for the clarification:)
I think for this delivery, the guide should focus on running the application in a Linux environment.
We haven't defined a requirement for this procedure to be applicable to Windows and MacOS environments, and rather than delivering a half-completed multi-environment workflow. I'd focus on delivering one that is complete and consistent for all runtimes on Linux.
Also, we are likely going to rework this procedure extensively for the next train, as we plan to make it work with an OpenShift cluster rather than localhost.

Ladicek: Ayea. I'd just revert to the previous wording.

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
```diff
@@ -0,0 +1,135 @@
...
+The default behavior of {Node}-based applications is to expose metrics at the `/metrics` endpoint.
+This is also what Prometheus expects.
+--
+
+* The Prometheus server running on `localhost` using Docker:
++
+--
+[source,bash,opts="nowrap"]
+----
+# docker run -it --rm --network host -p 9090:9090 -v $PWD/prometheus.yml:/etc/prometheus/prometheus.yml prom/prometheus
```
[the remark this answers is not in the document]
Ladicek: That certainly shouldn't be necessary and looks potentially dangerous.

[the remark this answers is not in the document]
inoxx03: @akoniar I have not encountered this issue when running the application locally. And I'd be inclined to agree with @Ladicek that using privileged mode might have its own negative security implications.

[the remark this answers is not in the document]
akoniar: @Ladicek , @inoxx03  Ok, I'll try to figure out why this line return me: level=error ts=2018-11-05T12:58:41.238321664Z caller=main.go:617 err="error loading config from \"/etc/prometheus/prometheus.yml\": couldn't load configuration (--config.file=\"/etc/prometheus/prometheus.yml\"): open /etc/prometheus/prometheus.yml: permission denied"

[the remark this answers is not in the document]
akoniar: @Ladicek @inoxx03 I have this problem because of SELinux. Thanks.

inoxx03:
@Ladicek  I remove the comment about  `--network host` and changed the wording back to the original one. The procedure now only covers running the apps on Linux. WDYT?

docs/topics/proc_exposing-application-metrics-using-prometheus-on-node-js.adoc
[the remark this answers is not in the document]
Ladicek: Hah, of course it's SELinux. We should figure out how to run with SELinux, as that's enabled by default in RHEL and CentOS and Fedora AFAIK.

Ladicek:
Yea, that part LGTM.

tradej:
retest this please

inoxx03:
@lance @danbev I've updated the steps to launch and configure Prometheus to make the procedure accessible for MacOS and Windows environments. The procedure seems to work locally, Please let me know what you think. Thank you.

inoxx03:
@tradej I added a `launcher-docs-only` variable and the corresponding conditionals around the metrics related content

tradej:
LGTM, merging.

PR #1207, opened 2018-11-12
RHOARDOC-1126: Description of JIRA fields and workflow

tradej:
Resolves RHOARDOC-1126

I have followed the common status meanings in [this document](https://mojo.redhat.com/docs/DOC-1174305).


tradej:
retest this please

tradej:
[test]

docs/topics/ref_rhoardoc-jira-fields.adoc
```diff
@@ -0,0 +1,61 @@
...
+Only valid for _Bugs, Enhancements_ and _Tasks._
+
+Due Date:: A specific date before which the issue should be resolved.
++
+If not set, check the linked epic or parent issue if any.
+They might have a Due Date set.
+
+Resolution:: How the issue was resolved.
++
+By default, it set to _Unresolved._
```
inoxx03: Should this be:

>  By default, it is set to _Unresolved._

?

docs/topics/ref_rhoardoc-jira-fields.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Description:: A complete description of the issue.
+
+Sprint:: Which sprint the issue belongs to.
++
+Only put an issue in a sprint if you are reasonably confident you can finish it within the sprint.
+
+Story Points:: The amount of effort the issue requires.
++
+Numbers from the Fibonacci sequence are used, each denoting an increasing workload: *1, 2, 3, 5, 8, 13.*
```
inoxx03: Maybe we could add a note here that the recommended number of total story points per writer per sprint is 13. Or are we not set on that yet? WDYT?

spisiakm: So if I get that right, when I create a JIRA issue with 13 story points for example, then that issue would span the whole sprint? :)

docs/topics/ref_rhoardoc-jira-fields.adoc
tradej: Good catch, thanks!

docs/topics/ref_rhoardoc-jira-fields.adoc
tradej: @inoxx03 Sure, I'll add it.

@spisiakm That's the recommendation, yes. Anything larger is probably an epic and should be broken down into smaller JIRAs.

tradej:
@ccopelloRH Good catch, thanks!

docs/topics/ref_rhoardoc-jira-fields.adoc
inoxx03: @spisiakm Yes, that might happen if the issue sin question is truly monolithic. Or, if it isn't, the issue might be split into smaller chunks that can be done over the course of several sprints.
Also, the 13 story points per sprint are recommended not prescribed, so more points are acceptable, if the writer can handle it.

spisiakm: @tradej @inoxx03 Thanks for clarification!

tradej:
@inoxx03 Thanks for the feedback, all should be fixed now and ready to merge.

PR #1208, opened 2018-11-13
RHOARDOC-1469: Prepare attributes file for portal publishing

tradej:
The `document-attributes-portal.adoc` file is intentionally left empty so that it can be filled in the downstream repository without conflicts, and to avoid a build warning in case it didn't exist.

The include directive is at the end of the file so that any values defined in the `document-attributes.adoc` file can be overwritten in the `document-attributes-portal.adoc` file.

Resolves RHOARDOC-1469

### DIFF
--- a/docs/topics/templates/document-attributes.adoc
+++ b/docs/topics/templates/document-attributes.adoc
@@ -214,3 +214,6 @@
 
 // Link S2I process (primarily for guides mission interaction sections)
 :link-s2i-process: https://docs.openshift.com/container-platform/latest/architecture/core_concepts/builds_and_image_streams.html#source-build
+
+include::document-attributes-portal.adoc[]
+


Ladicek:
Why don't we maintain `document-attributes-portal.adoc` here in upstream? There's nothing really secret in it, isn't it?

tradej:
@Ladicek, we need to maintain some differences in the code (at the very least undefine `:launcher-docs-only:` downstream), and I thought it would make more sense to merge changes with an empty file as opposed to one with content, as there should be no conflicts. The secrecy isn't really the aim of this change.

Ladicek:
OK, but you could, instead of including `-portal` at the end of the main file, include the main file at the beginning of `-portal`, right? That way, you could maintain everything in a single repo. That of course requires the downstream tooling to use `-portal` as the main file, which I'm not sure is easy/possible to do.

tradej:
@Ladicek, that requires all `master.adoc` files of the respective guides to include `document-attributes-portal.adoc` instead of `document-attributes.adoc` in the downstream repository. I'm not saying that's impossible or even hard, but I think it's a bit more error-prone and we'll still run (albeit a pretty low) risk of merge conflicts when committing to `master.adoc`.

Ladicek:
I see, so it's clearly not "easy/possible to do". I think I get it now, thanks.

inoxx03:
[test]

PR #1209, opened 2018-11-15
chore:  change the references of bucharest-gold to the new nodeshift …

lholmquist:
…and nodeshift-starters org


tradej:
add to whitelist

PR #1210, opened 2018-11-15
Fixed commands for interacting with the API in Vert.x and fixed in…

huntley:
Ok. Commit message. a little too long. Blah.

@inoxx03 @tradej @rdruss @cescoffier 
Updated commands for Vert.x and fixed syntax errors for Node.js. Also updated the examples to be more in line with each other and the actual data returned from the examples.

Please review and provide feedback. Thanks!

### DIFF
--- a/docs/topics/proc_interacting-with-the-crud-api.adoc
+++ b/docs/topics/proc_interacting-with-the-crud-api.adoc
@@ -45,13 +45,16 @@ $ curl http://{app-name}-{project-name}.{os-route-hostname}/api/fruits
 ----
 [ {
   "id" : 1,
-  "name" : "Cherry",
+  "name" : "Apple",
+  "stock" : 10
 }, {
   "id" : 2,
-  "name" : "Apple",
+  "name" : "Orange",
+  "stock" : 10
 }, {
   "id" : 3,
-  "name" : "Banana",
+  "name" : "Pear",
+  "stock" : 10
 } ]
 ----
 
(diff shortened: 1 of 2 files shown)


inoxx03:
[test]

docs/topics/proc_interacting-with-the-crud-api.adoc
```diff
@@ -18,7 +18,7 @@ When you have finished creating your application booster, you can interact with
 $ oc get route {app-name}
 ----
 
-[source,option="nowrap",subs="attributes+"]
+[source,bash,option="nowrap",subs="attributes+"]
```
tradej: This isn't a bash script, so we shouldn't mark it as such. Asciidoctor applies a stylesheet automatically to code block to highlight syntax etc., which wouldn't make sense in this case.

tradej:
@huntley, please include a link to the JIRA you address in the PR, which in this case I believe to be RHOARDOC-1461.

docs/topics/proc_interacting-with-the-crud-api.adoc
inoxx03: Agree with @tradej here.
It's possible we use these code block attributes  it in some places in our docs, but we should not be doing so, because of the syntax highlighting.
Instead, for terminal output examples in plain text, we could drop the `source,bash` attributes, which renders the inside the codeblock in plain monospace font only (no highlights).

huntley:
Addressed @tradej comments for https://issues.jboss.org/browse/RHOARDOC-1461

tradej:
Changed the name to get the WIP checker unstuck.

PR #1211, opened 2018-11-16
RHOARDOC-1472: Separated Launcher and Portal attribute definitions into separate files

tradej:
The attribute definitions are now in two separate files with no redefinitions taking place. This means that if an attribute is to be different in the Launcher and on the Portal, it should be defined in the `document-attributes-launcher.adoc` and `document-attributes-portal.adoc` files as opposed to the main `document-attributes.adoc` file. In that file, only truly common attributes go.

Which of the specific files shall be included is determined by the content of the `environment.adoc` file that only contains `:launcher-docs-only:` in this repository, and `:portal-docs-only:` downstream.

The specific files are included in the middle of the `document-attributes.adoc` file to make it possible to both derive specific attributes from common ones or vice versa (this is why #1208 didn't work).

I have updated the Contribution Guide section about the repository layout accordingly.

This is a rework of #1208, resolves RHOARDOC-1472.

### DIFF
--- a/docs/topics/con_repository-filesystem-layout.adoc
+++ b/docs/topics/con_repository-filesystem-layout.adoc
@@ -15,21 +15,24 @@ NOTE: The triple-dot indicates there are more files or directories in the partic
 │   │   ├── styles/ <4>
 │   │   ├── templates/ <5>
 │   │   │   ├── document-attributes.adoc <6>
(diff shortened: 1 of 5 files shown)


Ladicek:
This is a very good solution, nicely done!

tradej:
Thank you, @Ladicek :)

PR #1213, opened 2018-11-20
RHOARDOC-1478

inoxx03:


inoxx03:
[test]

inoxx03:
@tradej Thank you for the feedback.
Upon investigation, it seems that the both instances could be replaced by the variable.
I'll add the fix for the former to this PR, and for the latter, I'll file separate PRs in thorntail/thorntail and wildfly-swarm-prod/wildfly-swarm.

tradej:
OK, @inoxx03. Will you be including the Thorntail resync as part of this PR?

inoxx03:
@tradej If his is the only change  we're syncing to launcher docs form Thorntail, then I can include it in this PR. I'd also be fine with syncing in a separate PR, if preferable to you:)

tradej:
@inoxx03 I generally sync Thorntail sources in the same PR because the changes therein are an integral part to the issue's resolution, and merging just the changes here would result in a "broken" state of the docs. In this case, however, the change is more cosmetic in its nature, so in this case, either way is good by me.

No matter how you decide, can you please solicit a SB QE or Engg member review for this change? I'm not sure if the version change from `1.2` to `latest` is necessarily correct, and would like it validated.

inoxx03:
@tradej RE: sync?: Yes, that makes sense, I will sync the docs in this PR, then.
RE: ENG review, certainly, the version change is my concern, too, but SB ENG requested that the reference point to the latest available image version. I'll have them ack that nonetheless, just in case:)

inoxx03:
@tradej Thorntail files are synced. Please feel free to merge. Thank you.

PR #1217, opened 2018-11-22
RHOARDOC-1470: Refactored attributes

tradej:
I have done the following things:

* Consolidated the naming that it follows a namespace-like policy:
  * `name-$SOMETHING` for names, like `name-guide-vertx` or `name-launcher`.
  * `link-$SOMETHING` for links, like `link-guide-vertx` or `link-launcher-oso`.
  * `version-$something` for version numbers, like `version-vertx` or `version-spring-boot-maven-plugin`.
  * `value-$SOMETHING` for example values, like `value-name-project` or `value-url-oso-auth`.
* Left `red-hat*` attributes as they are for now for compatibility reasons with the Thorntail repo.
* Left product names with the capitalised notation for the same reason.
* Verified there are minimal differences between the existing and new rendered docs. The only differences I have created are the link texts in the `-resources` files as they were sometimes inconsistent.

Due to the scope of the changes, I recommend reviewing individual commits.

### DIFF
--- a/docs/contrib-guide/master.adoc
+++ b/docs/contrib-guide/master.adoc
@@ -20,7 +20,7 @@ _TBD_
 
 == Documentation
 
-This chapter contains information about contributing and releasing the {docs-name} documentation. You can also contribute to the documentation by filing an issue in our link:https://issues.jboss.org/projects/RHOARDOC/[Jira] project with corrections or feedback.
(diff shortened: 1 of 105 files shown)


inoxx03:
nice work @tradej :) Rendered docs and attribute changes in source both LGTM. Thank you:) merging

PR #1222, opened 2018-11-23
fixes RHOARDOC-1485

inoxx03:


Ladicek:
Both works, but I agree that using Maven properties for dependency versions is a "recommended practice". I'd keep `<version>${version.vertx}</version>` even if the value of `version.vertx` isn't shown in the example.

tradej:
If you don't mind, @inoxx03, I'd close this as _not a bug._

spisiakm:
In almost every Vert.x example (e.g. https://github.com/openshiftio-vertx-boosters/vertx-configmap-booster/blob/master/pom.xml#L29 ) we use `vertx.version`, so if we want to be more in line with the Vert.x dev flow, I'd suggest using that - reverting previous commit. :)

inoxx03:
@mspisiak good point. I think I went overboard on the changes here.
In fact, using the Maven variable reference would be my preferred way of doing this.
I can do so for both examples.
@tradej:  re: maven variable reference, for Vert.x it was indeed so, for the SB example I included AsciiDoc variable reference s instead, then decided to consolidate that for both runtimes to for consistency.

I'd be inclined to replace the explicit version substitutions in both examples with maven variable references.

I'll update the PR

inoxx03:
@mspisiak Thank you for the feedback, I've updated the PR to use Maven variable references in the `pom.xml` examples. Please let me know what you think.

@tradej It was not entirely a Maven variable reference, because the syntax was AsciiDoc. This was an oversight on my part. I have factored in the feedback form Martin and changed the reference to use the Maven-appropriate syntax (including the variable names). Please feel free to close as _not a bug_, if the proposed fix is unnecessary:)

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
```diff
@@ -22,7 +22,7 @@ Specify the `<type>pom</type>` and `<scope>import</scope>`:
       <dependency>
         <groupId>io.vertx</groupId>
         <artifactId>vertx-dependencies</artifactId>
-        <version>${vertx.version}</version>
+        <version>{version-vertx}</version>
```
spisiakm: @inoxx03 I was pointing mainly to this line in my comment - this change is not really necessary :)

spisiakm: Oh and btw you mentioned different guy previously :D

tradej:
@inoxx03 For the sake of consistency, I do agree we should use the Maven variable reference in the Spring Boot POMs as well. I suggest you undo all changes in this PR except the SB POM change and submit that. Also please make sure the variable is replaced in all the relevant places. Thank you.

P. S. This doesn't seem like an urgent problem to me, so please defer this in favour of more high-priority tasks if you have any.

inoxx03:
@tradej thank you for approving the PR, the penultimate forced update was a mistake and has been reverted. Please feel free to merge when you have a chance. Thank you:)

PR #1215, opened 2018-11-22
WIP: Initial pass of Prometheus configuration for Vertx.

huntley:
https://issues.jboss.org/browse/RHOARDOC-1442

@cescoffier @inoxx03 @tradej 

Please review the Java code in the pull request. @inoxx03 deserves the credit and I any blame.
To be consistent please provide an example similar to what is provided in the upstream Spring Boot example.

PR #1226, opened 2018-11-27
RHOARDOC-1442: Prometheus deployment for Vert.x (fixes, based on #1215)

tradej:


docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+. Include the `vertx-micrometer`, `micrometer-registry-prometheus` and `vertx-web` dependencies in the `pom.xml` file of your application:
++
+.pom.xml
+[source,xml]
+----
+<dependency>
+  <groupId>io.vertx</groupId>
+  <artifactId>vertx-micrometer-metrics</artifactId>
+  <version>3.5.4</version>
```
spisiakm: I don't know if we want to have Vert.x version hard-coded here. You'd need to update each time there's a release and it should come from `vertx-dependencies` anyway :)

tradej: Oh, I missed that, thanks. Does `${vertx.version}` work?

spisiakm: It should work by removing `<version></version>` tag altogether, as version is then automatically pulled/aligned from `vertx-dependencies` in project's dependency management section. How to set that up is described in 2. procedure step here https://launcher.fabric8.io/docs/vertx-runtime.html#creating-an-application_vertx

tradej: Ah, of course. Thank you, @spisiakm, I'll file a follow-up PR to remove the versions.

docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
```diff
@@ -0,0 +1,190 @@
...
+import io.vertx.core.http.HttpServerOptions;
+import io.vertx.micrometer.MicrometerMetricsOptions;
+import io.vertx.micrometer.VertxPrometheusOptions;
+import io.vertx.micrometer.backends.BackendRegistries;
+
+public class MyVertxApp extends AbstractVerticle {
+
+    @Override
+    public void start() {
+        Vertx vertx = Vertx.vertx(new VertxOptions().setMetricsOptions(
```
tsegismont: This is wrong. You should't create another Vert.x instance inside the verticle `start` method.

As discussed in the other PR, the right way to configure Micrometer metrics with 3.5.x is to:

- create your own `Launcher` class
- override the `beforeStartingVertx` method to setup metrics options
- override `afterStartingVertx` to configure the meter filter
- configure you launcher class in the POM file

docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
```diff
@@ -0,0 +1,190 @@
...
+----
+--
+
+. Set the `<vertx.verticle>` property in the `pom.xml` file of your application to your custom class:
++
+[source,xml,options="nowrap"]
+----
+<properties>
+  ...
+  <vertx.verticle>org.acme.MyVertxApp</vertx.verticle>
```
tsegismont: Also set the `vertx.launcher` property to point to your own `Launcher` class.

docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
tradej: Could you please provide a code sample? I think it will be the most efficient solution. Thank you.

tsegismont: @tradej here your go:

```java
public class CustomLauncher extends Launcher {

  @Override
  public void beforeStartingVertx(VertxOptions options) {
    options.setMetricsOptions(new MicrometerMetricsOptions()
      .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true)
        .setStartEmbeddedServer(true)
        .setEmbeddedServerOptions(new HttpServerOptions().setPort(8081))
        .setEmbeddedServerEndpoint("/metrics"))
      .setEnabled(true));
  }

  @Override
  public void afterStartingVertx(Vertx vertx) {
    PrometheusMeterRegistry registry = (PrometheusMeterRegistry) BackendRegistries.getDefaultNow();
    registry.config().meterFilter(
      new MeterFilter() {
        @Override
        public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
          return DistributionStatisticConfig.builder()
            .percentilesHistogram(true)
            .build()
            .merge(config);
        }
      });
  }
}
```

PR #1230, opened 2018-11-30
RHOARDOC-1446: Restricted Vert.x Prometheus procedure to Launcher only

tradej:


tradej:
Thanks, @inoxx03 @spisiakm.

PR #1232, opened 2018-12-05
Resolve comment for Vert.x Prommetheus procedure

huntley:
Resolves https://issues.jboss.org/browse/RHOARDOC-1492

Late breaking comment from @tsegimont.

### DIFF
--- a/docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
+++ b/docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
@@ -131,7 +131,7 @@ public class CustomLauncher extends Launcher {
 }
 ----
 
-. Create a custom `AbstractVerticle` class and override the `start` method to collect metrics.
+. Create a custom `Verticle` class and override the `start` method to collect metrics.
 For example, measure the execution time using the `Timer` class:
 +
 .Example CustomVertxApp.java file
@@ -159,9 +159,9 @@ public class CustomVertxApp extends AbstractVerticle {
 
     vertx.setPeriodic(1000, l -> {
       timer.record(() -> {
-      
+
         // Do something
-      
+
       });
     });
   }


docs/topics/proc_exposing-application-metrics-using-prometheus-with-vertx.adoc
```diff
@@ -159,9 +159,9 @@ public class CustomVertxApp extends AbstractVerticle {
 
     vertx.setPeriodic(1000, l -> {
       timer.record(() -> {
-
```
spisiakm: Is this line just indentation fix? :)

tradej: It seems to me like dangling white space.

huntley: Hmm. Ok. Excuse me for being dense but is the issue on line 162?

tradej: @huntley Basically just why you did it.

As a side note, it is considered a semi-bad practice to include unrelated changes like removing dangling whitespace in commits. Is it what happened? (You don't have to edit this PR it if so, just make sure you do white space removal in a separate PR, ideally for the entire doc set at the same time.)

huntley: Ahh. The spacing change was inadvertent. What is the easiest way to revert that change?

tradej: @huntley Don't worry about it, it's too much work and it's not really "forbidden." Unless you did that deliberately, what I suspect is that your editor's set up to remove dangling white space automatically. If that's the case, I'd only ask you to find that setting and turn it off.

PR #1235, opened 2018-12-11
Add Java arguments to Vert.x applications deployed using FMP

huntley:
Despite the misnamed branch, this resolves https://issues.jboss.org/browse/RHOARDOC-1439

@spisiakm @cescoffier 

Do we need to provide a rationale for why a developer should add these Java options for FMP deployments?

### DIFF
--- a/docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
+++ b/docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
@@ -33,5 +33,12 @@ endif::[]
 
 +
 In this profile, the Fabric8 Maven plugin is invoked for building and deploying the application to OpenShift.
-
+. Create a `deployment.yaml` file in the `src/main/fabric8` directory and add the `JAVA_OPTIONS` described in the example.
++
+ifdef::built-for-vertx[]
+[source,yaml,options="nowrap",subs="attributes+"]
+----
+include::resources/vert-x/vertx-fmp-deployment.yaml[]
+----
+endif::[]
 . Deploy the application to OpenShift according to instructions in xref:deploying-an-application-to-openshift_{context}[].
--- a/docs/topics/resources/vert-x/vertx-fmp-deployment.yaml
+++ b/docs/topics/resources/vert-x/vertx-fmp-deployment.yaml
@@ -1,0 +1,13 @@
+spec:
+  template:
+    spec:
+      containers:
+        - name: vertx
+          env:
+            - name: KUBERNETES_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.namespace
+            - name: JAVA_OPTIONS
+              value: '-Dvertx.cacheDirBase=/tmp -Dvertx.jgroups.config=default'


inoxx03:
[test]

docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
```diff
@@ -33,5 +33,12 @@ endif::[]
 
 +
 In this profile, the Fabric8 Maven plugin is invoked for building and deploying the application to OpenShift.
-
+. Create a [filename]`deployment.yaml` file in the [filename]`src/main/fabric8` directory and add the JAVA_OPTIONS shown in the following example.
```
inoxx03: Is there a particular reason you've opted to use semantic markup?
It is absolutely OK to do so, I'm just wondering.
(while this is 100% AsciiDoctor-compatible, I'm unaware of any tools  that would require it at the moment).

docs/topics/resources/vert-x/vertx-fmp-deployment.yaml
```diff
@@ -0,0 +1,15 @@
...
+          env:
+            - name: KUBERNETES_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.namespace
+            - name: JAVA_OPTIONS
+              value: '-Dvertx.cacheDirBase=/tmp -Dvertx.jgroups.config=default-configs/default-jgroups-kubernetes.xml -Djava.net.preferIPv4Stack=true'
+            - name: JAVA_ARGS
+              value: '-Dvertx.cacheDirBase=/tmp -Dvertx.jgroups.config=default-configs/default-jgroups-kubernetes.xml -Djava.net.preferIPv4Stack=true -cluster'
```
spisiakm: @cescoffier do we really need to duplicate this? From what I understood from [FMP documentation]( https://maven.fabric8.io/#generator-vertx), more specifically from here

> You can pass application parameter by setting the `JAVA_ARGS` env variable. You can pass system properties either using the same variable or using `JAVA_OPTIONS`. For instance, create src/main/fabric8/deployment.yml with the following content to configure `JAVA_ARGS`:

It looks like it should be enough to specify the java args and system properties in just `JAVA_ARGS`

docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
huntley: Mostly out of curiosity as I was looking to see if there would be a rendered difference for the the same markup. I tend to favor semantic distinctions as they can be important for L10N. I didn't see a distinction in my local build.

docs/topics/resources/vert-x/vertx-fmp-deployment.yaml
huntley: @spisiakm @cescoffier I would like to know the answer to this as well. I noted that both were in the example.

docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
tradej: I've been thinking about starting using semantic markup, and I'm very much open to the idea. RHEL docs do that, but many other products don't. Having said that, because we don't use semantic markup anywhere else in the doc set yet, I'd rather you didn't use it here.

docs/topics/proc_deploying-an-existing-application-to-openshift.adoc
```diff
@@ -33,5 +33,12 @@ endif::[]
 
 +
 In this profile, the Fabric8 Maven plugin is invoked for building and deploying the application to OpenShift.
-
+. Create a filename]`deployment.yaml` file in the `src/main/fabric8` directory and add the JAVA_OPTIONS and JAVA_ARGS shown in the example.
```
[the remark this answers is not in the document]
huntley: Thanks for the comments. I don't mind nitpicky reviews. Though citing of the Chicago Manual of Style can lead to burns from hot lead.

huntley:
Thanks for the comments. These are resolved.

docs/topics/resources/vert-x/vertx-fmp-deployment.yaml
huntley: Fixed!

tradej: I don't quite understand, does that mean that both of those declarations should be there?

huntley: Both arguments were in the code sample provided by @cescoffier. Until I hear otherwise, I am leaving them both.

tradej: @huntley, please seek confirmation from @cescoffier to make sure this is indeed correct. We need an authoritative answer from either engineering or QE before merging. Thanks.

cescoffier: Sorry, just had a look.

It should be:
```
            - name: JAVA_OPTIONS
              value: '-Dvertx.cacheDirBase=/tmp -Dvertx.jgroups.config=default-configs/default-jgroups-kubernetes.xml -Djava.net.preferIPv4Stack=true'
            - name: JAVA_ARGS
              value: '-cluster'
```

tradej: @cescoffier Thank you, much appreciated.

huntley:
Updated yams example to remove JAVA_ARGS and updated the module to remove mention of JAVA_ARGS.

tradej:
@huntley, can you please find out the supportability status of using Vert.x clustering (see https://github.com/fabric8-launcher/launcher-documentation/pull/1235#discussion_r245261134)? I want to make sure we don't publish something unsupported. Thank you.

huntley:
@tradej Seems to work much better when I remember to push my changes ....

inoxx03:
[test]

PR #1238, opened 2018-12-17
RHOARDOC-1495: Started validating the Thorntail guide

tradej:


tradej:
@Ladicek The emphasis in the verbatim text does not conform to the DocBook schema, which is used when rendering downstream documentation and therefore also when validating the XML output of Asciidoctor.

Of course, validating XML instead of AsciiDoc is not ideal (because the syntactic construction in question is indeed well-formed AsciiDoc), but since Asciidoctor has no useful validation and it is DocBook XML that is fed to the downstream renderer, we're going with it.

Ladicek:
Thanks for explanation!

PR #1239, opened 2018-12-18
Adding maven repository information to second location for Vert.x

huntley:
@tradej 

Please review

### DIFF
--- a/docs/topics/proc_configuring-your-application-to-use-spring-boot.adoc
+++ b/docs/topics/proc_configuring-your-application-to-use-spring-boot.adoc
@@ -1,7 +1,7 @@
 [id='configuring-your-application-to-use-spring-boot_{context}']
 = Configuring your application to use {SpringBoot}
 
-To configure your application to use {SpringBoot}:
+Reference the {SpringBoot} BOM (Bill of Materials) artifact in the `pom.xml` file at the root directory of your application.
 
 .Prerequisites
 
@@ -9,8 +9,7 @@ To configure your application to use {SpringBoot}:
 
 .Procedure
 
-. Reference the {SpringBoot} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Open the `pom.xml` file, add the `me.snowdrop:spring-boot-bom` artifact to the `<dependencyManagement>` section, and specify the `<type>pom</type>` and `<scope>import</scope>`:
 +
 --
 [source,xml,options="nowrap",subs="attributes+"]
@@ -31,9 +30,11 @@ Specify the `<type>pom</type>` and `<scope>import</scope>`:
   ...
 </project>
 ----
+--
 
-Include the following properties in your `pom.xml` file to track the version of {SpringBoot} and the {SpringBoot} Maven Plugin you are using:
-
+. Include the following properties to track the version of {SpringBoot} and the {SpringBoot} Maven Plugin you are using:
++
+--
 [source,xml,subs="attributes+",options="nowrap"]
 ----
 <project>
@@ -49,6 +50,7 @@ Include the following properties in your `pom.xml` file to track the version of 
 
 . Specify the repositories containing {ProductShortName} {SpringBoot} Starters and the {SpringBoot} Maven Plugin:
 +
+--
 [source,xml,subs="attributes+",options="nowrap"]
 ----
   <!-- Specify the repositories containing RHOAR artifacts. -->
@@ -69,9 +71,11 @@ Include the following properties in your `pom.xml` file to track the version of 
     </pluginRepository>
   </pluginRepositories>
 ----
+--
 
 . Reference `spring-boot-maven-plugin` as the plugin used to package your application:
 +
+--
 [source,xml,options="nowrap",subs="attributes+"]
 ----
 <project>
@@ -100,6 +104,7 @@ Include the following properties in your `pom.xml` file to track the version of 
   ...
 </project>
 ----
+--
 
 .Additional resources
 
--- a/docs/topics/proc_configuring-your-application-to-use-vertx.adoc
+++ b/docs/topics/proc_configuring-your-application-to-use-vertx.adoc
@@ -1,7 +1,7 @@
 [id='configuring-your-application-to-use-vertx_{context}']
 = Configuring your application to use {VertX}
 
-To configure your application to use {VertX}:
+Reference the {VertX} BOM (Bill of Materials) artifact in the `pom.xml` file at the root directory of your application.
 
 .Prerequisites
 
@@ -9,8 +9,7 @@ To configure your application to use {VertX}:
 
 .Procedure
 
-. Reference the {VertX} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Open the `pom.xml` file, add the `io.vertx:vertx-dependencies` artifact to the `<dependencyManagement>` section, and specify the `<type>pom</type>` and `<scope>import</scope>`:
 +
 --
 [source,xml,options="nowrap",subs="attributes+"]
@@ -31,9 +30,10 @@ Specify the `<type>pom</type>` and `<scope>import</scope>`:
   ...
 </project>
 ----
-
-Include the following properties in your `pom.xml` file to track the version of {VertX} and the Vert.x Maven Plugin you are using:
-
+--
+. Include the following properties to track the version of {VertX} and the Vert.x Maven Plugin you are using:
++
+--
 [source,xml,options="nowrap",subs="attributes+"]
 ----
 <project>
@@ -49,6 +49,7 @@ Include the following properties in your `pom.xml` file to track the version of 
 
 . Reference `vertx-maven-plugin` as the plugin used to package your application:
 +
+--
 [source,xml,options="nowrap",subs="attributes+"]
 ----
 <project>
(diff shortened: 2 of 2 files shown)


tradej:
add to whitelist

huntley:
Made a few editorial changes to smooth out the flow with the addition of another step.

huntley:
With regard to the introduction, I think mentioning the file that the following steps apply to is helpful. Previously, the file was mentioned in the first step and not the second. With the addition of a third step, I think this information needs to be in the introduction. I do see a mistake though, the <dependencyManagement> should be in the first step. Perhaps that addresses your second bullet but I can also change Specify to Add.

huntley:
@tradej Check out the latest changes. If they are ok, I will update proc_configuring-your-application-to-use-springboot as well.

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
```diff
@@ -1,16 +1,15 @@
...
 
 .Prerequisites
 
 * A Maven-based application
 
 .Procedure
 
-. Reference the {VertX} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Add `<type>pom</type>` and `<scope>import</scope>` to the `<dependencyManagement>` section of the `pom.xml` file:
```
tradej: This is better, but factually incorrect. You need to add the `io.vertx:vertx-dependencies` artifact that isn't there in the first place, and specify the `<type>` and `<scope>` along with it. @spisiakm, correct me if I'm wrong, please.

spisiakm: That's right. I would say something like

> Add `vertx-dependencies` dependency along with proper `scope` and `type` to the dependency management section of the pom.xml file.

huntley: @tradej @spisiakm I addressed your latest commits in my most recent commit. I will point out that the example pom.xml referenced later in the runtime guide does include the dependency management artifact.

huntley:
Made minor changes to the Spring Boot module and one change to Vert.x (plus improved line spacing). I did not modify the order of steps 3 &4 in the Spring Boot module (or could be done to Vert.x) to make the modules match. I can easily do this if requested but I'm not sure the order is important or that level of consistency is necessary and useful.

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
```diff
@@ -1,16 +1,15 @@
...
 
 .Prerequisites
 
 * A Maven-based application
 
 .Procedure
 
-. Reference the {VertX} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Add the `<dependencyManagement>` artifact and specify the `<type>pom</type>` and `<scope>import</scope>`:
```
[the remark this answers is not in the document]
spisiakm: It looks like we had a similar conversation before, [see above](https://github.com/fabric8-launcher/launcher-documentation/pull/1239#discussion_r245244400). Also the previous wording

> . Reference the {VertX} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.

seems perfectly fine to me, no need to change that. 

The `<dependencyManagement>` section in Maven's pom.xml file is a place where you specify dependency (or artifact) versions and other attributes that are used throughout the project (that is, a place where you centralize dependency information). See [official description.](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management)
 The `vertx-dependencies` Maven artifact (module, ..) is the BOM artifact that does version alignment of other artifacts. For example, it specifies a version of `vertx-web` were you to use it. 

I guess the only thing I would maybe add is a mention about specifying `<scope>` and `<type>`.

huntley:
Thanks @tradej & @spisiakm I modified the Vert.x wording about the dependencyManagement artifact based on the wording in the Spring Boot module. I though the SP module was more likely to be correct.

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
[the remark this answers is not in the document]
huntley: @mspiniak I am looking to resolve this PR but I want to make sre we are all in agreement here. The text you reference above was fine but its placement within the module was not. This text appeared in the first step of a procedure. All of the steps, which are independent of each other, describe edits to the pom.xml file and the first step is specific to <dependencyManagement>. As a result, I moved the text about the pom.xml to the introductory section of the module. All that said, do you have any comments or want any changes to the module in this PR? Thanks!.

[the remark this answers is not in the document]
tradej: @spisiakm ^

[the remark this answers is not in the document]
spisiakm: So I'm still trying to understand why you want to change the wording/steps description. I find [this wording](https://access.redhat.com/documentation/en-us/red_hat_openshift_application_runtimes/1/html-single/eclipse_vert.x_runtime_guide/#configuring-your-application-to-use-vertx_vertx) pretty clear and understandable. The only thing I might do is to remove this sentence

> To configure your application to use Eclipse Vert.x:

from below the heading, as the heading already says that.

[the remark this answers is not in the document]
huntley: @spisiakm Sorry about the typo in your handle in my prior post. Thanks to Sheldon for catching and correcting it.

I am not doing a good job of explaining my point so let me try again. The wording you refer to was fine fora simple two-step process. It is bad practice, however, to put information in one step that also applies to other steps in the procedure. The `pom.xml` applies to all steps in this module so I moved that to the top. The `<dependencyManagement>` is specific to step 1 so that remains in step 1. Removing mention `pom.xml` from the first step resulted in awkward working so I recast step 1 to make better sense. Perhaps in vain...:-)

docs/topics/proc_configuring-your-application-to-use-spring-boot.adoc
```diff
@@ -1,16 +1,15 @@
 [id='configuring-your-application-to-use-spring-boot_{context}']
 = Configuring your application to use {SpringBoot}
 
-To configure your application to use {SpringBoot}:
+Reference the `pom.xml` file in the {SpringBoot} BOM (Bill of Materials) artifact to configure your application.
```
tradej: I'm sorry, I don't know how this slipped past me. You reference the BOM in the `pom.xml` file, not the other way around.

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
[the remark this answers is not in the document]
tradej: @huntley I understand the reasoning, but I think in this case it leads to slightly worse user experience. As a former developer, I tend not to pay a lot of attention to introductions if I know this is indeed the file I'm looking for, and in this case, this means I only glance over it, and go directly to Prerequisites and step 1. That step does not mention where I should add the artifact, so even though I can find that information easily just a few lines above, it makes me stop and think for a second. If you want to adhere to the best practice, I'd say the step 1 should be _Open the `pom.xml` file in the root directory of your application._ or some such to improve the user experience.

docs/topics/proc_configuring-your-application-to-use-spring-boot.adoc
huntley: I fixed this. Let me know what you think.

docs/topics/proc_configuring-your-application-to-use-spring-boot.adoc
```diff
@@ -1,16 +1,15 @@
...
 
 .Prerequisites
 
 * A Maven-based application
 
 .Procedure
 
-. Reference the {SpringBoot} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Open the `pom.xml` file and add the `<dependencyManagement>` artifact  and specify the `<type>pom</type>` and `<scope>import</scope>`:
```
tradej: @huntley I'm still not sure whether the formulation _`<dependencyManagement>` artifact_ is semantically correct. Your opinion, @spisiakm?

spisiakm: It's not. Again, see https://github.com/fabric8-launcher/launcher-documentation/pull/1239#discussion_r246795747 . `<dependencyManagement>` is not an artifact, it's a section in a `pom.xml` file that does some specific stuff.

docs/topics/proc_configuring-your-application-to-use-vertx.adoc
```diff
@@ -1,16 +1,15 @@
...
 
 .Prerequisites
 
 * A Maven-based application
 
 .Procedure
 
-. Reference the {VertX} BOM (Bill of Materials) artifact in the `<dependencyManagement>` section of the `pom.xml` file of your application.
-Specify the `<type>pom</type>` and `<scope>import</scope>`:
+. Open the `pom.xml` file and add the `<dependencyManagement>` artifact and specify the `<type>pom</type>` and `<scope>import</scope>`:
```
spisiakm: Same as above, `<dependencyManagement>` is not an artifact.

huntley:
@tradej @spisiakm Thanks for you help. I've made Sheldon's changes.

spisiakm:
LGTM now

tradej:
Thank you, gentlemen, merged.

PR #1240, opened 2019-01-07
RHOARDOC-1474: Log in to OSL in the Minishift addon installation

tradej:


huntley:
Looks good to me. Thanks Sheldon.

PR #1247, opened 2019-03-13
RHOARDOC-1544: replace JDK version prerequisite with a variable.

inoxx03:


docs/topics/proc_creating-an-application.adoc
```diff
@@ -22,7 +22,7 @@ The following procedure shows you how to:
 .Prerequisites
 
 * Maven installed.
-* JDK 8 or later installed.
+* {prerequisite-jdk-version}
```
tradej: I think this is not the case with Thorntail 2.2.0, so we can't make this replacement. @huntley, please have a look into this, thanks.

inoxx03: @tradej, we can conditionalize it to show a Java 8 JDK only for TT. WDYT?

huntley:
I confirmed that Thorntail doesn't support JDK 11 until the 2.4 release at
the end of April.

*Huntley Eshenroder*

Customer Content Services

Red Hat

huntley@redhat.com     M: 708-224-9334

inoxx03:
@huntley Thank you for verifying the supported JDK version with the Thorntail team.
I will change the current variable to take a `parameter-version-jdk-supported` value that will be defined in the `master.adoc` file for each runtime, ensuring correct attribute value substitution in each runtime guide.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -19,8 +19,8 @@ Deploy your {parameter-runtime-name} application to Red Hat Enterprise Linux.
 * A Maven-based {parameter-runtime-name} application.
 * The Red Hat Maven repository link:{parameter-link-appdev-guide}[specified] in the `pom.xml` of your project.
 * RHEL 7 installed.
-* OpenJDK 8 installed.
-* Maven 3.5 or later installed.
+* {prerequisite-jdk-version}
+* Maven 3.6.0 or later installed.
```
Ladicek: Who requested this? :-)

inoxx03: No one:) I just got a bit ahead of myself :D let me fix this:)

tradej:
LGTM, merging.

PR #1253, opened 2019-03-25
Fixed Asciidoctor version to 1.5.8 to avoid build error

tradej:


inoxx03:
@tradej LGTM, mering :)

PR #1256, opened 2019-03-30
RHOARDOC-1531: Add procedure for deploying Java-based applications to standalone RHEL

inoxx03:


docs/topics/assembly_creating-a-basic-wildfly-swarm-application.adoc
```diff
@@ -8,6 +8,18 @@ In addition to xref:mission-rest-http-wf-swarm[using a booster], you can create
 
 include::thorntail/docs/howto/writing-an-application-from-scratch/index.adoc[leveloffset=+1]
 
+:parameter-runtime-name: {Thorntail}
+:parameter-group-name-runtime-plugin: io.thorntail
+:parameter-artifact-name-runtime-plugin: thorntail-maven-plugin
+:parameter-version-runtime-plugin: ${thorntail-maven-plugin.version}
```
Ladicek: I don't know about other runtimes, but in case of Thorntail, the Maven plugin version is identical to the Thorntail version itself. So this should be something like `${version.thorntail}` (I personally prefer the `version` string to be a prefix, not a suffix :-) ).

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,122 @@
...
+    </repository>
+  </repositories>
+  ...
+  <pluginRepositories>
+      <pluginRepository>
+        <id>{red-hat-ga-repo-id}</id>
+        <name>{red-hat-ga-repo-name}</name>
+        <url>{red-hat-ga-repo-url}</url>
+      </pluginRepository>
+    </pluginRepositories>
```
Ladicek: Indentation.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,122 @@
...
+              <goal>initialize</goal>
+              <goal>package</goal>
+            </goals>
+          </execution>
+        </executions>
+endif::[]
+ifdef::built-for-thorntail[]
+        <executions>
+          <execution>
+            <id>package</id>
```
Ladicek: For Thorntail, please remove the `<id>`.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,122 @@
...
+    </plugins>
+  </build>
+  ...
+</project>
+----
+
+. Package your application.
+Consider packaging your application as a fat jar to include all of the necessary dependencies.
+ifdef::built-for-vertx[The Vert.x Maven Plugin packages applications as fat jars link:https://reactiverse.io/vertx-maven-plugin/#introduction[by default].]
+ifdef::built-for-thorntail[The Thorntail Maven Plugin lets you package applications as link:https://docs.thorntail.io/2.3.0.Final/#uberjar_thorntail[_uberjars_] or link:https://docs.thorntail.io/2.3.0.Final/#creating-a-hollow-jar_thorntail[hollow JARs].]
```
Ladicek: Please link to 2.4.0.Final.

Ladicek: Also applications are _not_ packaged as hollow JARs. Best to not mention the hollow JAR at all in this context.

docs/topics/assembly_creating-a-basic-vertx-application.adoc
```diff
@@ -13,10 +13,24 @@ In addition to xref:mission-rest-http-vertx[using a booster], you can create new
...
 
 include::proc_deploying-an-application-to-openshift.adoc[leveloffset=+1]
+
+:parameter-runtime-name: {VertX}
+:parameter-group-name-runtime-plugin: io.reactiverse
+:parameter-artifact-name-runtime-plugin: vertx-maven-plugin
+:parameter-version-runtime-plugin: ${vertx-maven-plugin.version}
+:parameter-jar-file-name: my-app-fat.jar
+include::proc_deploying-an-application-to-standalone-rhel.adoc[leveloffset=+1]
+:parameter-jar-file-name:
```
tradej: These should be undefined, i. e. `:parameter-jar-file-name!:`, not just set to an empty string. Same elsewhere.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+//
+// Rationale: This procedure is the same for 2 or more runtimes.
+[id='deploying-an-application-to-standalone-rhel_{context}']
+= Deploying an application to standalone Red Hat Enterprise Linux
+
+You can deploy a {parameter-runtime-name} application to Red Hat Enterprise Linux.
+
+.Prerequisites
+
+* A Maven-based application.
```
tradej: I would add a mention of the runtime here to imply we expect they're already using it.

docs/topics/assembly_creating-a-basic-vertx-application.adoc
inoxx03: Yup, they certainly should be. My bad. Let me fix that. Thanks for the catch:)

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
inoxx03: Ok, let me add that, too

docs/topics/assembly_creating-a-basic-vertx-application.adoc
inoxx03: I have now undefined ```{parameter-jar-file-name}``` everywhere.

inoxx03:
@geoand @liborfuka Would you please verify if this is correct for Spring Boot?
@rdruss @akoniar Would you please verify that this is correct for Vert.X?

docs/topics/assembly_creating-a-basic-wildfly-swarm-application.adoc
geoand: Same thing for the Spring Boot plugin - the version of the plugin is the same as the version of Spring Boot you use

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+              <goal>package</goal>
+            </goals>
+          </execution>
+        </executions>
+endif::[]
+ifdef::built-for-spring-boot,built-for-spring-boot-1-5[]
+        <executions>
+          <execution>
+              <goals>
+               <goal>repackage</goal>
```
geoand: :+1:

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+  </build>
+  ...
+</project>
+----
+
+. Package your application.
+Consider packaging your application as a fat jar to include all of the necessary dependencies.
+ifdef::built-for-vertx[The Vert.x Maven Plugin packages applications as fat jars link:https://reactiverse.io/vertx-maven-plugin/#introduction[by default].]
+ifdef::built-for-thorntail[You can use the Thorntail Maven Plugin to package you application as an link:https://docs.thorntail.io/2.4.0.Final/#uberjar_thorntail[_uberjar_].]
+ifdef::built-for-spring-boot,built-for-spring-boot-1-5[You can use the Spring Boot Maven Plugin to repackage your target artifacts as link:https://docs.spring.io/spring-boot/docs/current/reference/html/build-tool-plugins-maven-plugin.html#build-tool-plugins-maven-packaging[executable JAR or WAR files].]
```
geoand: In the `jar` case, should we be calling it uberjar like for Thorntail?

inoxx03: based on the docs I've read, the terms _uberjar_ and _fat jar_ are used as equivalents. Vert.x, for example, call it _fat jar_, so both appear to be correct.

geoand: Indeed they are, but I don't see either of them being mentioned here for Spring Boot :P

docs/topics/assembly_creating-a-basic-wildfly-swarm-application.adoc
akoniar: Not in Vert.x case. The version of the plugin is different as the version of Vert.x.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+* A Maven-based {parameter-runtime-name} application.
+* RHEL 7 installed.
+* OpenJDK 8 installed.
+* Maven 3.5 or later installed.
+
+
+.Procedure
+
+. Ensure that you specify the link:https://maven.repository.redhat.com/[Red Hat Maven repository] in the `pom.xml` of your project.
```
akoniar: I think in Vert.x case this step is unnecessary because is pretty well described in the previous [Creating an application sub-section](https://github.com/fabric8-launcher/launcher-documentation/blob/749cdcc066149349825620c05b713e9130ca1e18/docs/topics/proc_creating-an-application.adocl), but if you want to keep it you should add this step to [Deploying an application to OpenShift sub-section](https://github.com/fabric8-launcher/launcher-documentation/blob/749cdcc066149349825620c05b713e9130ca1e18/docs/topics/proc_deploying-an-application-to-openshift.adoc) too.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean package
+----
+
+. Start your application:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ java -jar target/{parameter-jar-file-name}
```
akoniar: For Vert.x case in whole documentation is used for deploying in local machine `mvn vertx:run` so it should be a little bit confusing to use `mvn clean package && java -jar *.jar` command in only standalone case.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,121 @@
...
+.Prerequisites
+
+* A Maven-based {parameter-runtime-name} application.
+* RHEL 7 installed.
+* OpenJDK 8 installed.
+* Maven 3.5 or later installed.
+
+
+.Procedure
+
```
akoniar: I think the first step should be `cd myApp` as in Openshift case.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
akoniar: I think it is the same for other runtimes WDYT @Ladicek  @geoand  @liborfuka?

Ladicek: Definitely.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
inoxx03: @geoand I'll change it to _fat jar_ then:)

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
geoand: :+1:

inoxx03: @akoniar @geoand @Ladicek Thank you for pointing this out. When viewed in context of the other procedures on application development  that we have in the runtime guides, this step is indeed superfluous. I have therefore removed it and replaced with a prerequisite with a link pointing to the _Developing an application_ section that contains the POM with the repository configuration example. WDYT?

Ladicek: Prerequisite also makes sense, but I'd personally rather have all the important info in one place. But I don't mind either way.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,107 @@
...
+----
+$ cd myApp
+----
+
+. Ensure that the {parameter-runtime-name} Maven plugin is configured correctly in the `pom.xml` of your project:
++
+[source,xml,options="nowrap",subs="attributes+"]
+----
+<project ... >
+  ifdef::built-for-spring-boot,built-for-spring-boot-1-5[]
```
Ladicek: This doesn't work, it shows in the rendered text.

Ladicek:
I looked at the rendered form (good thing this Netlify thing!) and at least in the Thorntail context, this is totally useless. It's exactly the same as the _Creating an application from scratch_ chapter, except less detailed.

Ladicek:
However, at least in the Thorntail context, there's a way to actually make it useful. The _Creating an application from scratch_ chapter uses `mvn thorntail:run` to run the app, which is only good for testing, but not for production. So this chapter could say that an app from the previous chapter is a prerequisite, and then go on to say that `mvn package` is how the uberjar is built (and it resides at `target/myapp-thorntail.jar`), and `java -jar myapp-thorntail.jar` is how the app is supposed to be executed.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
geoand: Pasting so I don't go looking for it again: https://deploy-preview-1256--launcher-documentation.netlify.com/docs/spring-boot-runtime.html

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,107 @@
...
+        </executions>
+endif::[]
+      </plugin>
+    </plugins>
+  </build>
+  ...
+</project>
+----
+
+. Package and run your application:
```
akoniar: I pretty sure than mvn ${runtime}:run don't create a package.  It's is used only for the start app. Please check links 
[vertx-maven-plugin-doc](https://reactiverse.io/vertx-maven-plugin/#in-foreground-using-strong-a-href-vertx-run-vertx-run-a-strong)
[thorntail-maven-plugin-doc](https://docs.thorntail.io/2.2.1.Final/#using-thorntail-maven-plugin_thorntail#thorntail-maven-plugin-goals_thorntail)
[spring-boot-maven-plugin-doc](https://docs.spring.io/spring-boot/docs/current/maven-plugin/usage.html)

geoand: Good catch! I runs the application but I think calling it packacing would be a stretch :)

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,107 @@
...
+.Procedure
+
+. Navigate to the directory with your application project:
++
+[source,bash,subs="attributes+",options="nowrap"]
+----
+$ cd myApp
+----
+
+. Ensure that the {parameter-runtime-name} Maven plugin is configured correctly in the `pom.xml` of your project:
```
akoniar: Originally I wanted to remove this section because of its redundant information.
How I see the doc now:

1. Create app
...
1.2. Define pom.xml
1.3. Create a class
...
3. Deploying an application to standalone
...
3.2 Ensure if in defined pom.xml from 1. step is correctly set ${runtime}-maven-plugin

akoniar: @inoxx03 if you want to keep it like that so please add this step to openshift with fabric8-maven-plugin and ${runtime}-maven-plugin subsection too. WDYT @Ladicek and @geoand?

tradej:
It's true that if we use the `mvn $RUNTIME:run`, the usefulness of the procedure is low. @inoxx03, please revert back to the `mvn package` + `java -jar` deployment method and reconcile all other comments around it. This is the spec we'll be delivering this issue to.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
inoxx03: @Ladicek Thank you for spotting the rogue conditional. It is fixed now.

inoxx03:
@tradej Thank you. will do.

inoxx03:
@Ladicek @akoniar @liborfuka @geoand Thank you for the feedback. Per the decision by @tradej, I have reverted to using the `mvn clean package` and the `java -jar ...` commands, instead of the runtime plugin-specific ones. I have retained the `cd myApp` step as well as the parameterized prerequisite for the Maven repo configuration. These changes should ensure that the procedure:
* avoids unnecessary reduplication
* remains true to the original docs requirement,
*  and retains the usefulness of the earlier pre-edit version.

Please let me know what you think.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
akoniar: I agree to keep it like that.

cmoulliard:
Remark: For standalone spring project, this is not needed to refer about kubernetes liveness/readiness as mentioned here https://github.com/fabric8-launcher/launcher-documentation/blob/36a082098aaa26eca4eb508c2ccdc2bc6f2c865d/docs/topics/proc_creating-a-spring-boot-2-1-application.adoc

inoxx03:
@cmoulliard Thank you for the comment:) I think that since we're dealing with deployment to standalone RHEL, with no cluster, I don't think the liveness and readiness check configuration is really within the scope of this procedure.
The reason we have them in [_Chapter 5.1_|https://launcher.fabric8.io/docs/spring-boot-runtime.html#creating-a-basic-springboot-application_spring-boot] is because the intention is to deploy the app to OpenShift, as described in [_Section 5.1.2_|https://launcher.fabric8.io/docs/spring-boot-runtime.html#deploying-an-application-to-openshift_spring-boot].
For this procedure, where the app is intended to run locally, I'd be inclined not to add these links here.
Does that make sense?

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,115 @@
...
+[source,bash,subs="attributes+",options="nowrap"]
+----
+$ cd myApp
+----
+
+Ensure that you properly configure the {parameter-runtime-name} Maven plugin:
++
+[source,xml,options="nowrap",subs="attributes+"]
+----
+<project ... >
```
geoand: For some reason in rendered version I am seeing `+`: https://deploy-preview-1256--launcher-documentation.netlify.com/docs/spring-boot-runtime.html#creating-a-basic-springboot-application_spring-boot, in section `5.1.2` after `Ensure that you properly configure the Spring Boot Maven plugin:`

inoxx03: @geoand Thanks for the catch. My bad, as I deleted a line form the procedure and did not add the numbering back. Fixed it now.

geoand: Thanks!

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,115 @@
...
+. Package your application:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean package
+----
++
+Consider packaging your application as a fat jar to include all of the necessary dependencies.
+ifdef::built-for-vertx[The Vert.x Maven Plugin packages applications as fat jars link:https://reactiverse.io/vertx-maven-plugin/#introduction[by default].]
+ifdef::built-for-thorntail[You can use the Thorntail Maven Plugin to package you application as an link:https://docs.thorntail.io/2.4.0.Final/#uberjar_thorntail[_uberjar_].]
```
tradej: The link should point to the uberjar module in our runtime guide. I am not sure about the other runtimes, but those links may be OK.

inoxx03: Thank you for the comment. Thee Spring Boot and Vert.x runtime plugins are not productized, and therefore the linked instructions have no counterpart in the Launcher or Product docs. I did, however update the Thorntail link so that now it points to Launcher docs.

tradej: I've just checked with Ali Heslin, and she says that the links to the other runtimes' doc sets in this context are perfectly acceptable.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,115 @@
...
+----
+
+. Package your application:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean package
+----
++
+Consider packaging your application as a fat jar to include all of the necessary dependencies.
```
Ladicek: With Thorntail, this happens always, as this is our primary delivery mechanism. "Considering it" sounds weird. I suspect other runtims are very similar in this regard.

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,115 @@
...
+ifdef::built-for-vertx[The Vert.x Maven Plugin packages applications as fat jars link:https://reactiverse.io/vertx-maven-plugin/#introduction[by default].]
+ifdef::built-for-thorntail[You can use the Thorntail Maven Plugin to package you application as an link:{link-guide-thorntail}#creating-an-uberjar_wf-swarm[_uberjar_].]
```
Ladicek: It's not that "you can use" the Thorntail Maven plugin to create an uberjar. You simply use the Thorntail Maven plugin to create an uberjar.

Ladicek:
Otherwise looking good, but from Thorntail perspective, this really doesn't add much value and I can see it merged with the previous chapter very easily.

Ladicek:
One more thing: deploying an app on standalone RHEL typically _doesn't_ involve creating the uberjar, at least not on the target system. The target production system usually won't have Maven installed at all. So deploying it means taking the uberjar and running it. Not much to talk about, unless we want to go into details such as tuning Java heap or GC etc. :-)

inoxx03:
@Ladicek Thank you, I see your point:) And I'm not opposed to merging the two procedures later, but for the time being, I'd be inclined to follow the docs requirement and deliver this as a separate procedure:)

inoxx03:
> The target production system usually won't have Maven installed at all. So deploying it means taking the uberjar and running it. Not much to talk about, unless we want to go into details such as tuning Java heap or GC etc. :-)

@Ladicek  I don't think that would be within the scope of this documentation requirement:) It does, however, seem like something that should have been recognized when the requirement being formulated. We won't be able to add the information here before we publish, but we can include it in a subsequent update:)

Ladicek:
What I wanted to say is that I think `mvn clean package` isn't exactly part of "deploying the app to standalone RHEL", it's a previous step. But I'm probably thinking too much about it :-)

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
inoxx03: @Ladicek Thank you for pointing this out. I have amended the wording to remove the confusing implication of a choice:)

docs/topics/proc_deploying-an-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+. Package your application:
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ mvn clean package
+----
++
+ifdef::built-for-vertx[The Vert.x Maven Plugin packages applications as fat jars link:https://reactiverse.io/vertx-maven-plugin/#introduction[by default].]
+ifdef::built-for-thorntail[The Thorntail Maven Plugin packages you application as an link:{link-guide-thorntail}#creating-an-uberjar_wf-swarm[_uberjar_].]
```
tradej: Packages _your_ application

inoxx03: @tradej thank you for catching that. Fixed.

tradej:
Thanks, everybody, for the feedback. As time is of the essence, I am merging this JIRA, while the outstanding feedback for improving the procedure is filed in [RHOARDOC-1587](RHOARDOC-1587) and will be addressed as time permits.

PR #1257, opened 2019-04-03
RHOARDOC-1531: Deployment procedure for Node.js apps on standalone RHEL.

inoxx03:


docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+.Procedure
+
+. Navigate to the directory with your application project:
++
+[source,bash,subs="attributes+",options="nowrap"]
+----
+$ cd myApp
+----
+
```
lholmquist: Might want to add something about install dependecies, with `npm install`

inoxx03: Thanks, added:)

tradej:
@lholmquist, can you please verify that the PR is acceptable to merge from the technical perspective? Thank you.

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,45 @@
...
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ npm install
+----
+
+. Start your application.
++
+[source,bash,options="nowrap",subs="attributes+"]
+----
+$ node app.js
```
lance: Depending on the application, there could be a number of different ways to start a Node app. This method assumes the entry point to the application is `app.js`. That may or may not be the case. It could be `index.js` or `www/app.js` or any other thing. Typically, though, best practice would be to have a start script in your package.json. In that case, the command would be `npm start`.

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,45 @@
...
+Server running at http://localhost:8080
+----
+
+.Verification steps
+
+. Use `curl` or your browser to verify your application is running at `http://localhost:8080`
```
lance: Again, this makes an assumption about the application being deployed using port 8080 for the web service.

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
```diff
@@ -0,0 +1,45 @@
+[id='deploying-a-node-js-application-to-standalone-rhel_{context}']
+= Deploying a {Node} application to standalone Red Hat Enterprise Linux
+
+Deploy a {Node} application to Red Hat Enterprise Linux.
+
+.Prerequisites
+
+* A {Node} application.
+* npm 6.4.1 installed
```
lance: This document does not specify a version for Node.js, so why are we specifying one for npm? We produce RPMs for both of them, and the versioning typically reflects a pair of binaries.

inoxx03: Yo're right It does look inconsistent this way. The required npm version I also documented in the release notes, so IMHO it doesn't have to be listed here, too. we could just say 

> npm installed

as the prerequisite.

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
inoxx03: Good point. Somehow I completely overlooked this.
The port configuration should be shown in the `app.js` example.

inoxx03:
@lance Thank you for your feedback. I would like to incorporate the changes that you are suggesting into this procedure. 
Per instructions from @tradej , this procedure must be delivered ASAP, preferably today, to still make in into the Train G  docs release. We are therefore requested to deliver the procedure as a Minimum Viable Product, preferably today.
Therefore, supposing we could initially tolerate some level of assumptions about our app in the procedure (such as the entry point being set to `localhost:8080`, the entrypoint being `app.js`) what would be the necessary changes to make this a  minimum viable procedure? We will, of course, make follow up edits ensuring the procedure is eventually complete. Apologies if this sounds like an unusual request, but the expedited merging is mostly a contningency measure.

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
inoxx03: I think it would be preferable to have an example of both `package json` and `app.js` showing where thew entrypoint and starter script are set.
(Due to time constraints we are currently facing resulting form the imminent closing of train G, I'm not sure we're able to deliver all of these changes int his PR, but I would definitely like to make them in a follow up PR, if possible.)

tradej:
@lance, thank you for the feedback, but time is of the essence at the moment, so I am merging based on @lholmquist's ack. @inoxx03 has filed a follow-up in [RHOARDOC-1588](https://issues.jboss.org/browse/RHOARDOC-1588), where we'll address all your points.

PR #1258, opened 2019-04-03
RHOARDOC-1583: Add tomcat starter as a dependency in the SB WAR file deployment procedure.

inoxx03:


docs/topics/proc_deploying-a-springboot-application-using-war-files.adoc
```diff
@@ -30,13 +30,34 @@ You must configure your build and deployment settings to ensure that your applic
...
 .Example `pom.xml`
 [source,xml,options="nowrap",subs="attributes+"]
 ----
+<project ...>
+  ...
+  <dependencies>
+    ...
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-tomcat</artifactId>
```
inoxx03: @liborfuka Would adding `spring-boot-starter-tomcat` as a dependency be  the correct way of specifying an application server?
Or would it be better to specify the productized JWS Tomcat artifact directly?

liborfuka: Its correct way to add tomcat as dependecny.

inoxx03: @liborfuka thank you

tradej:
LGTM, merging.

PR #1259, opened 2019-04-03
RHOARDOC-1578: Spring Boot 1.5.19. version bump

inoxx03:


docs/spring-boot-1-5-runtime/master.adoc
```diff
@@ -6,9 +6,9 @@ include::topics/templates/document-attributes.adoc[]
 
 //Legacy Spring Boot Version Attributes
 //override latest SB version
-:version-spring-boot: 1.5.17.RELEASE
-:version-spring-boot-bom: 1.5.17.SP1-redhat-00002
-:version-spring-boot-maven-plugin: 1.5.17.RELEASE
+:version-spring-boot: 1.5.19.RELEASE
+:version-spring-boot-bom: 1.5.19.Final-redhat-00001
+:version-spring-boot-maven-plugin: {version-spring-boot}
```
gytis: Shouldn't this be
```suggestion
:version-spring-boot-maven-plugin: 1.5.19.RELEASE
```

tradej: @gytis `{version-spring-boot}` is a variable reference, which will resolve to 1.5.19 at build time. Is that all right?

gytis: Sure, but why isn't it used also in `version-spring-boot:`? Those versions always align. I'm of course unfamiliar with the process, so if there is a reason for that then just go ahead and merge the PR

tradej: `:version-spring-boot: 1.5.19.RELEASE` is a variable definition, where you also set the variable to `1.5.19.RELEASE`, and `{version-spring-boot}` is a reference. Or dereference? It's been a long time since I've last programmed anything.

tradej: Here's a super-quick introduction to Asciidoctor variables (they're called _attributes_ for some reason): https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#attributes-and-substitutions

PR #1266, opened 2019-04-22
RHOARDOC-1594: Updates to README content

huntley:
https://issues.jboss.org/browse/RHOARDOC-1594

@tradej I have a couple of questions regarding these files.
- where is {mission} parameter defined? Would like to update that.
- the URLs in the http-secured modules (https://launcher.fabric8.io/docs/spring-boot-runtime.html#running-the-secured-booster-integration-tests_context) don't work and the new IDs won't help matters. Is adding an anchor to the corresponding runtime modules the best way to fix these links?


tradej:
> where is {mission} parameter defined? Would like to update that.

@gastaldi, do you have an answer to that, please?

> the URLs in the http-secured modules (https://launcher.fabric8.io/docs/spring-boot-runtime.html#running-the-secured-booster-integration-tests_context) don't work and the new IDs won't help matters. Is adding an anchor to the corresponding runtime modules the best way to fix these links?

@huntley Fix the URL in the `integrationTestSection` in the `*.properties` files in the `docs/topics/readme` directory.

@gastaldi, do you see a problem with the above approach?

gastaldi:
@tradej LGTM, good job :)

tradej:
Thanks for the review @gastaldi. Do you have an answer where the `{mission}` attribute is defined when the READMEs are rendered, please?

gastaldi:
@tradej the `{mission}` parameter is set when the Readme is rendered here: https://github.com/fabric8-launcher/launcher-backend/blob/master/core/core-impl/src/main/java/io/fabric8/launcher/core/impl/preparers/ReadmePreparer.java#L55

It uses what's defined in the `missions.name` in https://github.com/fabric8-launcher/launcher-booster-catalog/blob/master/metadata.yaml

tradej:
Thanks, @gastaldi. @huntley, leave the `{mission}` attributes as they are, then.

huntley:
@theashiot @nchauda @inoxx03 
Can you please provide a peer review and approve?

docs/topics/readme/configmap-README.adoc
```diff
@@ -1,13 +1,13 @@
-= ${mission} - ${runtime} Booster
+= ${mission} - ${runtime} Example Application
 
-IMPORTANT: While you can run and interact with this booster on localhost, you cannot take advantage of using externalized configuration with this booster without using OpenShift. For more details on using this booster with a single-node OpenShift cluster, CI/CD deployments, as well as the rest of the runtime, see the link:${guideURL}[${runtime} Runtime Guide].
+IMPORTANT: While you can run and interact with this example application on localhost, you cannot take advantage of using externalized configuration with this example application without using OpenShift. For more details on using this example application with a Single-node OpenShift cluster, CI/CD deployments, as well as the rest of the runtime, see the link:${guideURL}[${runtime} Runtime Guide].
```
snelluli1: Single-node OpenShift cluster --> Single-node OpenShift Cluster
Not sure why the attributes are not used for this in the README content. Writing the name manually across 100-odd files might result in such discrepancies. Consider using the attributes if time permits.

docs/topics/readme/osio-circuit-breaker-README.adoc
```diff
@@ -1,35 +1,35 @@
-= ${mission} - ${runtime} Booster
+= ${mission} - ${runtime} Example Application
 
-== Interacting with the Booster in Stage
-From your workspace, click _Stage - 1.0.1_ under _Applications_. This will open the stage version of your booster that you can interact with. For more details on interacting with this booster, see the link:${guideURL}[${runtime} Runtime Guide].
+== Interacting with the Example Application in Stage
+From your workspace, click _Stage - 1.0.1_ under _Applications_. This will open the stage version of your example application that you can interact with. For more details on interacting with this example application, see the link:${guideURL}[${runtime} Runtime Guide].
 
-NOTE: Deploying your booster to Stage may take a few minutes to complete.
+NOTE: Deploying your example application to Stage may take a few minutes to complete.
```
snelluli1: Deploying your example application to Stage--> Deploying your example application to the stage server?

docs/topics/readme/osio-crud-README.adoc
```diff
@@ -1,35 +1,35 @@
-= ${mission} - ${runtime} Booster
+= ${mission} - ${runtime} Example Application
 
-== Interacting with the Booster in Stage
-From your workspace, click _Stage - 1.0.1_ under _Applications_. This will open the stage version of your booster that you can interact with. For more details on interacting with this booster, see the link:${guideURL}[${runtime} Runtime Guide].
+== Interacting with the Example Application in Stage
+From your workspace, click _Stage - 1.0.1_ under _Applications_. This will open the stage version of your example application that you can interact with. For more details on interacting with this example application, see the link:${guideURL}[${runtime} Runtime Guide].
 
-NOTE: Deploying your booster to Stage may take a few minutes to complete.
+NOTE: Deploying your example application to Stage may take a few minutes to complete.
```
snelluli1: Stage--> the stage server?

docs/topics/readme/osio-crud-README.adoc
```diff
@@ -1,35 +1,35 @@
...
 * Click _Create_ at the top of the page.
 * Click _Create Workspace_.
 * Click _Open_.
 +
 NOTE: This will create a popup window so you may have to enable popups in your browser.
 
-* Make a change to your booster (e.g. edit `${fileLocation}`).
-* Click the _Run_ button (the triangle in the top next to the _Debug_ button) and choose _Run_ to build and deploy your booster in a temporary Che environment. Use the _Preview_ URL at the top of the log window to verify that your changes took effect.
+* Make a change to your example application (e.g. edit `${fileLocation}`).
+* Click the _Run_ button (the triangle in the top next to the _Debug_ button) and choose _Run_ to build and deploy your example application in a temporary Che environment. Use the _Preview_ URL at the top of the log window to verify that your changes took effect.
```
snelluli1: Two style-related suggestions that are beyond the scope of the Jira. Incorporate them only if time permits:
1. Don't use `e.g.` Use `for example` instead.
2. `the triangle in the top next to the _Debug_ button`: Where exactly on is this button located? "top" is not a good description enough. You could say "upper left" or "upper right".
`

docs/topics/readme/rest-http-README.adoc
```diff
@@ -17,9 +17,9 @@ $ cd ${targetRepository}
 $ ${localRunCMD}
 ----
 
-== Interacting with the Booster Locally
+== Interacting with the Example Application Locally
 
-To interact with your booster while it's running locally, use the form at `http://localhost:8080` or the `curl` command:
+To interact with your example application while it's running locally, use the form at `http://localhost:8080` or the `curl` command:
```
snelluli1: while it's running locally--> while it is running locally

docs/topics/readme/rest-http-secured-README.adoc
```diff
@@ -26,9 +26,9 @@ $ oc create -f service.sso.yaml
 $ ${localRunCMD}
 ----
 
-== Interacting with the Booster on a Single-node OpenShift Cluster
+== Interacting with the Example Application on a Single-node OpenShift Cluster
 
-To interact with your booster while it's running on a Single-node OpenShift Cluster, you first need to obtain the secured booster endpoint URL, which is the _${value-name-app}_ service:
+To interact with your example application while it's running on a Single-node OpenShift Cluster, you first need to obtain the secured example application endpoint URL, which is the _${value-name-app}_ service:
```
snelluli1: Same here: while it's running locally--> while it is running locally

docs/topics/readme/zip-rest-http-secured-spring-boot.properties
```diff
@@ -2,5 +2,5 @@ localRunCMD=mvn clean fabric8:deploy -Popenshift -DskipTests -DSSO_AUTH_SERVER_U
 app-name=sb
 gitCMD=git clone git@github.com:${loggedUser}/${targetRepository}
 guideURL=http://launcher.fabric8.io/docs/spring-boot-runtime.html
-prerequisite=IMPORTANT: This booster requires Java 8 JDK or greater and Maven 3.3.x or greater.
-integrationTestSection=To run the integration test on the Secured HTTP API booster follow the steps in the link:https://launcher.fabric8.io/docs/spring-boot-runtime.html#running-the-secured-booster-integration-tests_context[Secured Booster documentation].
+prerequisite=IMPORTANT: This example application requires Java 8 JDK or greater and Maven 3.3.x or greater.
+integrationTestSection=To run the integration test on the Secured HTTP API example application follow the steps in the link:https://launcher.fabric8.io/docs/spring-boot-runtime.html#running-the-secured-booster-integration-tests_context[Secured Booster documentation].
```
snelluli1: To run the integration test on the Secured HTTP API example application follow--> Need a comma between application and follow
`To run the integration test on the Secured HTTP API example application, follow`
There are a lot of instances with the same statement. Consider making this update in all instances for better clarity.

huntley:
Thanks @snelluli1. I will address your last comment in the very near future. I need to find that triangle myself :-)

docs/topics/readme/configmap-README.adoc
inoxx03: @snelluli1 thank you for pointing this out. The reason why don't substitute some terms using attributes in the README files is because these files were originally designed to be partially procedurally generated by the Launcher for the dual purpose of being displayed on-screen when creating an example application, as well as to be included in the ZIP filed with packaged example application projects. Because processing the README files with AsciiDoctor was not feasible in this case, we opted to use a custom BASH script that substituted entire source lines using shell variables.

inoxx03: @snelluli1 @huntley The above being said, I'd like to point out that we should no longer use the term _Single -node OpenShift Cluster_. We will be replacing it with _OpenShift cluster_ for [NON-PUBLIC PRODUCT RELATED REASONS]

docs/topics/readme/osio-circuit-breaker-README.adoc
inoxx03: @snelluli1 @huntley Good point, but as far as I know _Stage_ here does not refer to a stage server _per se_, but rather to a virtual environment for containerized application lifecycle management that was originally part of OpenShift.io. It has since been changed several times and I don't think we're currently using any of the OSIO READMEs

tradej:
The single-node OpenShift cluster was beyond the scope of this JIRA, merging as all other things were addressed.

PR #1272, opened 2019-04-26
WIP: Running on standalone RHEL 8

huntley:
https://issues.jboss.org/browse/RHOARDOC-1531

Apologies for jamming everyone into a single PR.

We are adding support for RHEL 8 to the runtime guides. This initial commit is to determine the changes between running RHOAR on RHEL7 and RHEL8. This leads to a few questions:

- Are there differences between how an example application is created?
- Are there differences between running an example application on RHEL7 and RHEL8? Perhaps we do not need a separate module?

Note that the runtimes guides are undergoing a restructure to improve the presentation of this information.


tradej:
ok to test

huntley:
Separate module is not necessary so closing this PR. Will submit new PRs with updated modules.

PR #1275, opened 2019-05-01
Fix broken cross references

huntley:
https://issues.jboss.org/browse/RHOARDOC-1620

@tradej I believe this commit fixes the broken cross references. The errors between the guides were all the same. If they aren't all fixed, I can address it in the morning my local time.


tradej:
ok to test

tradej:
@huntley I've fixed the remaining errors and I'm merging.

PR #1278, opened 2019-05-02
Added RHEL 8 as supported platform for Node.js

huntley:
https://issues.jboss.org/browse/RHOARDOC-1626

### DIFF
--- a/docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
+++ b/docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
@@ -5,9 +5,9 @@ Deploy a {Node} application to Red Hat Enterprise Linux.
 
 .Prerequisites
 
-* A {Node} application.
+* A {Node} application listening on port 8080.
 * npm 6.4.1 installed
-* RHEL 7 installed.
+* RHEL 7 or RHEL 8 installed.
 * {Node} installed
 
 


tradej:
ok to test

docs/topics/proc_deploying-a-node-js-application-to-standalone-rhel.adoc
```diff
@@ -5,9 +5,9 @@ Deploy a {Node} application to Red Hat Enterprise Linux.
 
 .Prerequisites
 
-* A {Node} application.
+* A {Node} application listening on port 8080.
 * npm 6.4.1 installed
-* RHEL 7 installed.
+* RHEL 7 or RHEL 8 installed.
 * {Node} installed
```
Ladicek: I don't have a say in this, but the order here looks strange. I'd expect it would go either from lowest-level to highest-level, or the opposite. Currently the order seems random. I'd use:

- RHEL 7 or RHEL 8 installed
- Node installed
- npm installed (actually it doesn't make too much sense, because npm is always installed as part of Node, if I understand correctly)
- Node application

tradej: That's a good point, @Ladicek, thanks. I'd rather keep this issue out of the scope of this PR, so I've filed [RHOARDOC-1639](https://issues.jboss.org/browse/RHOARDOC-1639) for all procedures.

tradej:
Because there was a refactoring that replaced these files, closing in favour of #1281.

PR #1293, opened 2019-06-11
RHOARDOC-1571: Removed micrometer-registry-prometheus dependency

huntley:
With the upcoming release, users no longer need to add micrometer-registry-prometheus as a dependency to the pom.xml file of their application.


tradej:
ok to test

inoxx03:
Merging...

PR #1304, opened 2019-07-10
RHOARDOC-1662: Replaced openshift/postgresql-92-centos7 

huntley:
with centos/postgresql-10-centos7 for CRUD example application


inoxx03:
[test]

PR #1313, opened 2019-07-30
RHOARDOC-1669 add missing instruction for installing on OpenShift

chrira:
Installation does not work without this customization of the custom resource.

### DIFF
--- a/docs/topics/proc_installing-launcher-operator.adoc
+++ b/docs/topics/proc_installing-launcher-operator.adoc
@@ -60,24 +60,6 @@ You must replace these value with the {name-launcher} frontend route URL when yo
 $ oc create secret generic launcher-oauth-github --from-literal=clientId=LAUNCHER_GITHUB_OAUTH_CLIENT_ID --from-literal=secret=LAUNCHER_GITHUB_OAUTH_CLIENT_SECRET
 ----
 
-.. Edit the custom resource definition YAML for your {name-launcher} instance to specify your OpenShift Web Console URL.
-+
-.launcher_cr.yaml
-[source,yaml,options="nowrap",subs="attributes+"]
-----
-apiVersion: launcher.fabric8.io/v1alpha2
-kind: Launcher
-metadata:
-  name: launcher
-spec:
-# OpenShift Configuration
-  openshift:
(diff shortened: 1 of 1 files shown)


chrira:
> @chrira Thanks for the feedback! I have a few comments on the wording. Let me know what you think.

I do not find your comments. Where do I have to look at them?

huntley:
Now I don't see them either. I will recreate.

docs/topics/proc_installing-launcher-operator.adoc
```diff
@@ -126,7 +126,7 @@ Based on the templates, OpenShift automatically installs:
 * an access role definition for the Launcher operator
 * access role bindings for the Launcher operator
 
-. Install the `launcher_cr.yaml` custom resource to initiate the deployment of the {name-launcher} tool.
+. Customize the `launcher_cr.yaml` custom resource with your OpenShift Console URL and install it in order to initiate the deployment of the {name-launcher} tool.
```
huntley: I think this works better as two separate steps: modify the custom resource and then install it. Something like:

8. Edit the 'launcher_cr.yml' custom resource to add your OpenShift Console URL:

####### OpenShift Configuration
  openshift:
    consoleUrl: #<the OpenShift Console Url (ex: 'https://192.168.64.4:8443')>#

9. Install the 'launcher_cy.yaml' custom resource to deploy the {name-launcher} tool.

huntley:
@chrira Ok. Sorry about that. Comments are back. I have also asked one of our QE folks to take a look as well.

docs/topics/proc_installing-launcher-operator.adoc
dsimansk: I do agree with Huntley here. I'd like a bit more two steps approach with an example snippet.

chrira:
Thank you for the review. I changed the file according to your input.
I found a note on updating the launcher_cr.yaml ahead in the document. But it was before the download instruction of the ZIP file, which includes the launcher_cr.yaml file.
I moved that part.

huntley:
[test]

inoxx03:
[test]

PR #1320, opened 2019-08-08
RHOARDOC-1693: Fix cross references in Health Check example

huntley:
Updated link to OpenShift (points to RH docs instead of community too) and removed link to Kubernetes API as I am not sure of its usefulness and pointed to older version of API.


inoxx03:
[test]

PR #1337, opened 2019-09-13
RHOARDOC-1727: Vert.x 3.8.1 version bump

inoxx03:
* updated the release version string
* updated the Vert.x Maven Plugin version


inoxx03:
Thank you for the ACK @akoniar :)

PR #1342, opened 2019-10-08
fixed variable in output text

TimurSarbassov:
Name of the attribute :launcher-name: doesn't exist and was changed to existing :name-launcher:


inoxx03:
[test]

